PR #20, opened 2022-07-05
Overhauled error handling system

fbeutin-ledger:


src/ragger/backend/interface.py
```diff
@@ -66,44 +59,41 @@ def last_async_response(self) -> Optional[RAPDU]:
...
+    @abstractmethod
+    def __enter__(self) -> "BackendInterface":
+        raise NotImplementedError
+
+    @abstractmethod
+    def __exit__(self, exc_type: Optional[Type[BaseException]], exc_val: Optional[BaseException],
+                 exc_tb: Optional[TracebackType]):
+        raise NotImplementedError
+
+    def get_raise_policy(self) -> RaisePolicy:
```
lpascal-ledger, line 71: Getter and setter in Python are usually done through properties:
```python
@property 
def raise_policy(self): ...
@property.setter
def raise_policy(self, value): ...
```
And then used as a plain field:
``` python
backend.raise_policy = RaisePolicy.RAISE_NOTHING
```

In this case, as nothing is checked or processed, I'm not sure adding an abstraction is worth it. (?)

src/ragger/backend/speculos.py
```diff
@@ -85,18 +86,21 @@ def send_raw(self, data: bytes = b"") -> None:
...
         return result
 
-    @manage_error
+    @raise_policy_enforcer
     def exchange_raw(self, data: bytes = b"") -> RAPDU:
         logger.debug("Sending '%s'", data.hex())
         return RAPDU(0x9000, self._client._apdu_exchange(data))
 
+    @raise_policy_enforcer
+    def get_last_async_response(self, response) -> RAPDU:
```
lpascal-ledger, line 101: I'm not sure this method should be "public", it seems to be purely internal to this class. If it is (internal), prefix it with a `_` to explicit this should not be used from outside.

src/ragger/backend/speculos.py
```diff
@@ -105,12 +109,7 @@ def exchange_async_raw(self, data: bytes = b"") -> Generator[None, None, None]:
                                                p2=data[3],
                                                data=data[5:]) as response:
             yield
-            try:
-                self._last_async_response = RAPDU(0x9000, response.receive())
-            except ApduException as error:
-                if self.raises and not self.is_valid(error.sw):
-                    raise self._error(error.sw, error.data)
-                self._last_async_response = RAPDU(error.sw, error.data)
+            self._last_async_response = self.get_last_async_response(response)
```
lpascal-ledger, line 112: ... following previous comment about `get_last_async_response`, if you don't want to create a specific fonction to trigger the decorator, you can use it directly as a fonction:
```python
self._last_async_response = RAPDU(0x9000, raise_policy_enforcer(response.receive)())
```

tests/backend/test_ledgercomm.py
```diff
@@ -56,24 +57,35 @@ def test_receive_ok(self):
                 rapdu = self.backend.receive()
         self.check_rapdu(rapdu, payload=payload)
 
+    def test_receive_ok_raise(self):
+        failure, payload = 0x8000, b"something"
```
lpascal-ledger, line 61: You could even use `0x9000` as it is consider the universal success status.

tests/backend/test_ledgercomm.py
```diff
@@ -56,24 +57,35 @@ def test_receive_ok(self):
...
         with patch("ledgercomm.transport.HID") as mock:
             self.hid = mock
             self.hid().recv.return_value = (failure, payload)
             with self.backend:
+                self.backend.set_raise_policy(RaisePolicy.RAISE_NOTHING)
                 rapdu = self.backend.receive()
         self.check_rapdu(rapdu, status=failure, payload=payload)
 
-    def test_receive_nok_raises(self):
+    def test_receive_nok_raise(self):
```
lpascal-ledger, line 81: To clarify why there was a 's' at `raises`, the tests is named as:
- `test_` - alright it's a test
- `receive` - the checked method name or member
- `nok` - we're testing a non-nominal behavior
- `raises` - the expected behavior: the method **raises** - like in `self.assertRaises` (most time it is not a verb, in this case it is)

As a coding habit it is not better or worse than any other habit, but I find unnecessary to have these cosmetic changes in a PR dealing with other stuff.

tests/test_boilerplate.py
```diff
@@ -2,11 +2,13 @@
...
 
-from ragger import ApplicationError, RAPDU
+from ragger import ExceptionRAPDU, RAPDU
+from ragger.backend.interface import RaisePolicy
 
 
-def test_error_returns_not_raises(client_no_raise):
-    result = client_no_raise.exchange(0x01, 0x00)
+def test_error_returns_not_raises(client):
+    client.set_raise_policy(RaisePolicy.RAISE_NOTHING)
```
lpascal-ledger, line 10: This feels better than before. More explicit, more flexible.

lpascal-ledger (dismissed):
Nice change. Some cosmetic comments.

src/ragger/backend/interface.py
fbeutin-ledger: I removed the abstraction and made raise_policy a plain attribute

src/ragger/backend/speculos.py
fbeutin-ledger: I did not know that
I think however that the method with the @ decorator is more clear, but I'll change if you prefer calling the decorator as a function

PR #22, opened 2022-07-27
Update speculos version

lpascal-ledger:


lpascal-ledger:
Fluffed the previous deployment...

PR #33, opened 2022-10-25
Feature/version

lpascal-ledger:
Adding firmware version checks using `semver` + corner cases (when using versions such as `2.1` which are expected from `speculos`).
This feature will help managing UX behavior changing depending on the SDK version.

### DIFF
--- a/src/ragger/firmware/versions.py
+++ b/src/ragger/firmware/versions.py
@@ -1,0 +1,85 @@
+"""
+   Copyright 2022 Ledger SAS
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+"""
+from enum import Enum
+from typing import Dict, Optional, Type
+
+from semver import VersionInfo
+
+
+class VersionManager(Enum):
+
+    @classmethod
+    def get_last(cls,
+                 major: Optional[int] = None,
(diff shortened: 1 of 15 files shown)


tests/backend/test_interface.py
```diff
@@ -46,11 +47,14 @@ def receive(self):
     def send_raw(self, *args, **kwargs):
         self.mock.send_raw(*args, **kwargs)
 
+    def finger_touch(self, *args, **kwargs):
```
agrojean-ledger, line 50: Why add finger touch in this PR ? @lpascal-ledger

lpascal-ledger: Glitching

src/ragger/firmware/versions.py
```diff
@@ -0,0 +1,80 @@
...
+        if major is not None and minor is None:
+            return max(v.value for v in cls if v.value.major == major)
+        if major is not None and minor is not None and patch is None:
+            return max(v.value for v in cls if (v.value.major == major and v.value.minor == minor))
+        try:
+            return max(
+                v.value for v in cls
+                if (v.value.major == major and v.value.minor == minor and v.value.patch == patch))
+        except ValueError:
+            raise ValueError(f"No version {major}.{minor}.{patch} found")
```
xchapron-ledger, line 46: Why catching here and not above? Maybe you could let the default message?

xchapron-ledger (commented):
You are fixing some `flake8` issue. Can you split them in an separate commit?
Also can you enforce the CI flake8 check? It's currently disbaled by `|| echo 1`?
If you do change the `.github/workflows/fast-checks.yml` can you fix the triggers? In my opinion, it would be better to have:
```
on:
  push:
    branches:
      - develop
      - master
  pull_request:
```
so that we don't have double tests on classical PR.

xchapron-ledger (approved):

src/ragger/firmware/versions.py
lpascal-ledger: Without management, the raised error would be `ValueError: max() arg is an empty sequence` which is not very helpfull.
But you're right this should be crafted for this case, but also with major only, or major + minor.

PR #34, opened 2022-11-02
Add ragger navigator features

xchapron-ledger:


src/ragger/backend/interface.py
```diff
@@ -287,6 +293,24 @@ def both_click(self) -> None:
...
+    def navigate(self, instruction: NavigationInstruction) -> None:
+        """
+        Performs a click on buttons of the device depending of the navigation instruction.
+
+        :raises NotImplementedError: If the navigation instruction is not implemented.
+
+        :return: None
+        :rtype: NoneType
+        """
+        if instruction == NavigationInstruction.GO_TO_NEXT_SCREEN:
```
agrojean-ledger, line 305: I don't understand why is this code in the interface definition

xchapron-ledger: because it is common to every backend? At least as long as we keep `right_click`, `left_click` and `both_click`?

src/ragger/backend/speculos.py
```diff
@@ -177,27 +183,27 @@ def navigate_until_snap(self,
                             start_img_idx: int = 0,
                             last_img_idx: int = 0,
                             take_snaps: bool = True,
-                            timeout: int = 30,
+                            timeout: int = 5,
```
agrojean-ledger, line 186: 5 seconds is too short a timeout if there are many screens in a flow, it will result in tests failing when they should succeed

src/ragger/backend/speculos.py
```diff
@@ -139,34 +149,30 @@ def _screenshot_equal_with_crop(self, path: Path, bytes: BytesIO, crop: Crop = N
...
             screenshot = self._client.get_screenshot()
-            if self._screenshot_equal_with_crop(path, BytesIO(screenshot), crop):
+            if self._snap_equal_with_crop(path, BytesIO(screenshot), crop):
                 return True
             now = time()
         return False
 
     def _compare_snaps(self, path: Path, test_case_name: Path, last_img_idx: int = 0) -> bool:
 
-        snapshot_golden_path = path / "snapshots" / self._firmware.device / test_case_name
```
agrojean-ledger, line 157: Why the hassle of renaming these variables just to make them a bit shorter ?

src/ragger/backend/interface.py
```diff
@@ -359,3 +383,28 @@ def navigate_until_snap(self,
         :rtype: NoneType
         """
         raise NotImplementedError
+
+    @abstractmethod
+    def navigate_and_compare(self, path: Path, test_case_name: Path,
+                             instructions: List[NavigationInstruction]) -> bool:
+        """
+        Navigate and compare snapshots.
```
agrojean-ledger, line 391: Make it explicit that there are as many navigation steps as provided by the instructions parameter

src/ragger/backend/speculos.py
xchapron-ledger: the idea is that we have `screenshot`, `snap` and `snapshot` naming, I think it would be better to harmonize on one name?

README.md
```diff
@@ -65,8 +65,10 @@ The `src/ragger/backend/interface.py` file describes the methods that can be imp
 * `right_click`: perform a right click on a device.
 * `left_click`: perform a left click on a device.
 * `both_click`: perform a click on both buttons (left + right) of a device.
+* `navigate`: performs a click on buttons of the device depending of the navigation instruction.
```
agrojean-ledger, line 68: I would reword it a bit : 

`navigate` : navigate on the device according to a set of navigation instructions provided.
`navigate and compare` : same as the previous method but take screenshots at every step and compare them with "golden images" after navigation is performed.

tests/backend/test_speculos.py
```diff
@@ -172,6 +173,22 @@ def test_navigate_and_compare_until_snap(self):
                         ROOT_SCREENSHOT_PATH, "generic", 0, 2)
                 self.assertTrue(ret)
 
+    def test_navigate_and_compare(self):
```
agrojean-ledger, line 176: Not necessary to test non-nominal cases ?

src/ragger/backend/interface.py
agrojean-ledger, line 305: @lpascal-ledger do you think this should be implemented in this way (in the interface directly) ?

src/ragger/backend/speculos.py
lpascal-ledger: `timeout` could be a one-image parameter, which would be extended when another screenshot is encountered. That would bring more flexibility.

agrojean-ledger: Not sure what you mean :thinking:

xchapron-ledger: That makes very long `ragger` test, but ok. I'm switching back to 30sec.

tests/backend/test_speculos.py
xchapron-ledger: tests added :)

src/ragger/backend/interface.py
agrojean-ledger: @xchapron-ledger actually I think the navigation method should be given a list of instructions and loop through them. Additionnaly, like navigate until snap , a "takesnaps" boolean parameter should be given to decide whether or not to take snapshots or not at every step.

Then , navigate and compare would just pass its own navigation instructions to navigate.

src/ragger/backend/speculos.py
```diff
@@ -248,3 +250,20 @@ def navigate_and_compare_until_snap(self,
...
         # Then compare all snapshots taken to golden images.
         return self._compare_snaps(path, test_case_name, img_idx)
+
+    def navigate_and_compare(self, path: Path, test_case_name: Path,
+                             instructions: List[NavigationInstruction]) -> bool:
+        snaps_tmp_path = self._get_snaps_dir_path(path, test_case_name, False)
+        snaps_tmp_path.mkdir(parents=True, exist_ok=True)
+
+        # First navigate to the last step and take snapshots of every screen in the flow.
+        for idx, instruction in enumerate(instructions):
```
agrojean-ledger, line 260: The navigate method should rather have a take snap parameter that would be set to true when called from navigate_and_compare rather than have a loop here and pass each instruction to another loop in navigate...

xchapron-ledger: Humf, I'm not found of this one. This would need to pass many parameters to the `navigate` method (snaps dir path, test case name, snap index, ...).
I rather go back and allow only one navigation instruction per `navigate` call.
What do you think?

agrojean-ledger: Then you might as well make the navigate method private if it's only used for one navigation instruction (the user can directly use "click_right" "click_left" and so on, I think).

src/ragger/backend/interface.py
```diff
@@ -29,6 +29,12 @@ class RaisePolicy(Enum):
     RAISE_ALL = auto()
 
 
+class NavigationInstruction(Enum):
```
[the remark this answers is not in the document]
lpascal-ledger: Given navigation will probably be a thing on its own, I'd say it could deserve its own module.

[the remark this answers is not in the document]
lpascal-ledger: But currently it is entangled with one of the concrete method of the interface, so it would make sense _for now_ to keep it inside the `backend` package.

src/ragger/backend/interface.py
agrojean-ledger: Ok

src/ragger/backend/interface.py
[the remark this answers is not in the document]
agrojean-ledger: @xchapron-ledger ?

agrojean-ledger (dismissed):

src/ragger/backend/interface.py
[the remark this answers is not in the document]
agrojean-ledger: Ok, I close this discussion because the future refactoring will probably throw this out of the windows.

src/ragger/backend/interface.py
```diff
@@ -362,3 +390,29 @@ def navigate_until_snap(self,
         :rtype: NoneType
         """
         raise NotImplementedError
+
+    @abstractmethod
+    def navigate_and_compare(self, path: Path, test_case_name: Path,
```
fbeutin-ledger, line 395: En plusieurs lignes c'est plus lisible

fbeutin-ledger (dismissed):

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,288 @@
...
+        :type middle_instruction_wait: float
+        :param last_instruction_wait: Sleeping time after the last navigation instruction
+        :type last_instruction_wait: float
+
+        :raises ValueError: If one of the snapshots do not match.
+
+        :return: None
+        :rtype: NoneType
+        """
+        if not isinstance(self._backend, SpeculosBackend):
```
lpascal-ledger, line 159: Not found of this. The whole point of the backends is to avoid these discriminated treatments. if we start splitting the usage depending on the backend type, the code will soon become illegible condition hell.
I'd rather accept that code run uselessly - but won't crash, all methods should be defined and correct - on some backend. **For now**. At some point, we will need those empty functions and replace them with more meaningful code - like ask a user to check if some golden image fits what he's seeing on the physical device screen.

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,288 @@
...
+        :type crop_first: Crop
+        :param crop_last: Crop (left, upper, right or lower pixels) last snapshot image for comparison
+                          (useful if using a generic snapshot).
+        :type crop_last: Crop
+
+        :return: img_idx
+        :rtype: int
+        """
+
+        if not isinstance(self._backend, SpeculosBackend):
```
lpascal-ledger, line 235: Same remark as in previous method.

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,288 @@
...
+            # When not using Speculos backend, taking snapshots is not possible
+            # therefore comparison is not possible too.
+            # TODO request user to interact with the device.
+            return 0
+
+        snaps_golden_path = self._get_snaps_dir_path(path, test_case_name, True)
+        if not snaps_golden_path.is_dir():
```
lpascal-ledger, line 242: Same remark as before. It's reasonnable to ask `self._get_snaps_dir_path` to assert if the directory exists or not (and is a directory). It could be optionnable through a keyword argument if we don't want to check everytime.

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,288 @@
...
+        # It saves time in non-nominal cases where the navigation flow does not start.
+        if self._compare_snap_with_timeout(first_golden_snap, timeout_s=2, crop=crop_first):
+            start = time()
+            # Navigate until the last snapshot specified in argument is found.
+            while not self._compare_snap_with_timeout(
+                    last_golden_snap, timeout_s=0.5, crop=crop_last):
+                now = time()
+                # Global navigation loop timeout in case the snapshot is never found.
+                if (now - start > timeout):
+                    raise ValueError(f"Timeout waiting for snap {last_golden_snap}")
```
lpascal-ledger, line 260: A `TimeoutError` would be more appropriate.

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,288 @@
...
+            # Validation action when last snapshot is found.
+            self.navigate([validation_instruction])
+
+            # Make sure there is a screen update after the final action.
+            start = time()
+            last_screen_update_timeout = 2
+            while self._compare_snap_with_timeout(last_golden_snap, timeout_s=0.5, crop=crop_last):
+                now = time()
+                if (now - start > last_screen_update_timeout):
+                    raise ValueError(
```
lpascal-ledger, line 283: `TimeoutError`?

src/ragger/navigator/nano.py
```diff
@@ -0,0 +1,52 @@
...
+from ragger.backend import BackendInterface
+from ragger.firmware import Firmware
+from ragger.navigator import NavigatorInterface, NavigationInstruction
+
+
+class NanoNavigator(NavigatorInterface):
+
+    _ARGS_KEY = 'args'
+
+    def __init__(self, backend: BackendInterface, firmware: Firmware):
```
lpascal-ledger, line 27: I think this `__init__` is superfluous, as it just redirect the argument to it motherclass, in the exact same order.

src/ragger/navigator/nano.py
```diff
@@ -0,0 +1,52 @@
...
+
+        :param instructions: Set of navigation instructions.
+        :type instructions: List[NavigationInstruction]
+
+        :raises NotImplementedError: If the navigation instruction is not implemented.
+
+        :return: None
+        :rtype: NoneType
+        """
+        for instruction in instructions:
```
lpascal-ledger, line 42: One way to uncouple the data from the code flow, in order to avoid future very long `if ... elif ... else ...` structure with potential redundant conditions we could miss, would be to create a dictionnary `{ <instruction> : <callback> }` (or whatever structure), then have a code like:
```python
for instruction in instructions:
    if not instruction in CALLBACKS:
        raise NotImplementedError
    CALLBACKS[instruction]()
```

src/ragger/utils/misc.py
```diff
@@ -0,0 +1,41 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+"""
+from typing import Optional, Tuple
+from pathlib import Path
+
+
+def app_path_from_app_name(app_dir, app_name: str, device: str) -> Path:
```
lpascal-ledger, line 20: `app_dir: Path`

lpascal-ledger (changes requested):

src/ragger/backend/interface.py
```diff
@@ -289,76 +289,11 @@ def both_click(self) -> None:
         raise NotImplementedError
 
     @abstractmethod
-    def navigate_and_compare_until_snap(self,
-                                        path: Path,
-                                        test_case_name: Path,
-                                        start_img_idx: int = 0,
-                                        last_img_idx: int = 0) -> bool:
+    def take_snapshot(self) -> Image:
```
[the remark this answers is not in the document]
lpascal-ledger: (or `io.BytesIO`)

src/ragger/navigator/nano.py
lpascal-ledger: (this could be onlined with a raison `lambda` but it would be very unclear)

lpascal-ledger: Advantage of having a dedicated `{ <instruction>: <callback> }` structure is that it is very easy to register (add/modify) custom behavior on any instruction.

src/ragger/backend/interface.py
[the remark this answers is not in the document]
xchapron-ledger: Ok, I'm changing to io.BytesIO

src/ragger/navigator/nano.py
xchapron-ledger: What if the instruction needs parameters? Let's say the `WAIT_1SEC` I've just added calls time.sleep(1). Shoud I create a `def sleep1(): time.sleep(1)`?

xchapron-ledger: Or maybe can you think of a way to add parameters? Maybe instructions could be not just an enum, but a class containing an enum and args?

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,288 @@
...
+    def _get_snaps_dir_path(self, path: Path, test_case_name: Path, is_golden: bool) -> Path:
+        if is_golden:
+            subdir = "snapshots"
+        else:
+            subdir = "snapshots-tmp"
+        return path / subdir / self._firmware.device / test_case_name
+
+    def _init_snaps_temp_dir(self, path: Path, test_case_name: Path) -> Path:
+        snaps_tmp_path = self._get_snaps_dir_path(path, test_case_name, False)
+        if snaps_tmp_path.exists():
```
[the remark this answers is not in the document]
xchapron-ledger: it will raise an error I guess. I don't think this is an issue as I don't think it will happen? And If not, the error should be quite understandable?

README.md
```diff
@@ -79,68 +83,161 @@ following fixtures:
 
 import pytest
```
fbeutin-ledger, line 84: Maybe put the conftest template in a dedicated file and refer to it in the README ?
Just a suggestion, I'm not even sure it would be better

README.md
```diff
@@ -79,68 +83,161 @@ following fixtures:
...
+
+@pytest.fixture(scope="session")
+def display(pytestconfig):
+    return pytestconfig.getoption("display")
+
+
 @pytest.fixture
-def firmware():
-    return NANOS_FIRMWARE
+def test_name(request):
```
fbeutin-ledger, line 122: Maybe make it a property ?

Even make it two properties:
self.test_name => return request.node.name
self.test_name_no_suffix => return self.test_name.split("[")[0]

src/ragger/navigator/interface.py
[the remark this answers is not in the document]
fbeutin-ledger: `snaps_tmp_path.isdir()` will ensure it exists + is a directory

src/ragger/navigator/interface.py
fbeutin-ledger: +1
And Navigation DOES make sense in LedgerComm and LedgerWallet, even if it's just asking the tester `Press right button once` and `Ensure that this image corresponds to the current Nano screen`

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        if not isinstance(self._backend, SpeculosBackend):
+            # When not using Speculos backend, taking snapshots is not possible
+            # therefore comparison is not possible too and eventually this method
+            # fallback to navigate one.
+            return self.navigate(instructions)
+
+        snaps_tmp_path = self._init_snaps_temp_dir(path, test_case_name)
+
+        sleep(first_instruction_wait)
```
fbeutin-ledger, line 172: Perfectly fine for a first step but it would be nice to eventually wait an event from speculos stating it is finished displaying
Maybe add a todo ?

src/ragger/utils/__init__.py
```diff
@@ -15,5 +15,11 @@
...
 from .packing import pack_APDU
+from .path import BtcDerivationPathFormat, pack_derivation_path
+from .path import bitcoin_pack_derivation_path
+from .misc import app_path_from_app_name, prefix_with_len
+from .misc import create_currency_config
 
-__all__ = ("RAPDU", "pack_APDU", "Crop")
+__all__ = ("RAPDU", "pack_APDU", "Crop", "BtcDerivationPathFormat", "pack_derivation_path",
+           "bitcoin_pack_derivation_path", "app_path_from_app_name", "prefix_with_len",
+           "create_currency_config")
```
fbeutin-ledger, line 25: Ok this effectively make the individual utils files redundant, good

src/ragger/navigator/nano.py
xchapron-ledger: Done!

src/ragger/navigator/interface.py
xchapron-ledger: yeah, could be nice to catch events for this...
I will add a todo for now.

README.md
xchapron-ledger: Done

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,288 @@
...
+            return screenshot_equal(f"{path}", snap)
+
+    def _compare_snap_with_timeout(self,
+                                   path: Path,
+                                   timeout_s: float = 5.0,
+                                   crop: Crop = None) -> bool:
+        start = time()
+        now = start
+        while not (now - start > timeout_s):
+            snap = self._backend.take_snapshot()
```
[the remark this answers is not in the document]
xchapron-ledger: `BackendInterface.take_snapshot` returns a `BytesIO` in the current version (I changed it I don't know exactly when).

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,288 @@
...
+        if not snaps_golden_path.is_dir():
+            raise ValueError(f"Golden snapshots directory ({snaps_golden_path}) does not exist.")
+
+        if not snaps_tmp_path.is_dir():
+            raise ValueError(f"Temporary snapshots directory ({snaps_golden_path}) does not exist.")
+
+        for i in range(0, last_img_idx + 1):
+            golden = self._get_snap_path(snaps_golden_path, i)
+            tmp = self._get_snap_path(snaps_tmp_path, i)
+            if not screenshot_equal(str(golden), str(tmp)):
```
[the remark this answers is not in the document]
xchapron-ledger: it is located in Speculos, so I don't want to change this.
But it already handles either path as str and BytesIO.

src/ragger/navigator/interface.py
xchapron-ledger: OR !

As discussed with @agrojean-ledger we could:
- add a `parser.addoption("--golden_snapshot")`
- pass it in `SpeculosBackend` instantiation `kwargs`
- have a specific behavior in theses cases which instead of comparing the snapshot it saves it in the golden dir.

with such behavior, we could go back to comparing the snapshot after each instruction, which would allow to do it on `LedgerComm` and `LedgerWallet`.

fbeutin-ledger: I suggest we postpone this to another PR, we'll see along the way

src/ragger/navigator/interface.py
xchapron-ledger: Discussed above

src/ragger/navigator/interface.py
xchapron-ledger: I've update the PR with an additional commit on which we can hopefully find a consensus!

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,297 @@
...
+
+class NavIns:
+
+    def __init__(self, id: NavInsID, args=(), kwargs: Dict[str, Any] = {}):
+        self.id = id
+        self.args = args
+        self.kwargs = kwargs
+
+
+class NavigatorInterface(ABC):
```
agrojean-ledger, line 44: @xchapron-ledger I think it shouldn't be called "Interface" as this class implements stuff. It's just a parent class to a device navigator... So you might just call it "Navigator" and rename the file `interface.py` to `navigator.py`

src/ragger/navigator/interface.py
```diff
@@ -0,0 +1,297 @@
...
+                if (now - start > last_screen_update_timeout):
+                    raise TimeoutError(
+                        f"Timeout waiting for screen change after last snapshot : {last_golden_snap}"
+                    )
+        else:
+            raise ValueError(f"Could not find first snapshot {first_golden_snap}")
+        return img_idx
+
+
+class NanoNavigator(NavigatorInterface):
```
agrojean-ledger, line 288: I would have had a file per device child class

src/ragger/navigator/interface.py
xchapron-ledger: done

lpascal-ledger (approved):

lpascal-ledger (approved):

PR #40, opened 2022-11-14
Unchain the doc

lpascal-ledger:


.github/workflows/documentation.yml
```diff
@@ -33,7 +39,7 @@ jobs:
     name: Deploy the documentation on Github pages
     runs-on: ubuntu-latest
     needs: generate
-    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+    if: github.event_name == 'push'
```
fbeutin-ledger, line 42: Why remove the master check ?

README.md
```diff
@@ -123,3 +127,17 @@ pytest --backend [speculos|ledgercomm|ledgerwallet] <tests/path>  # will run tes
 
 The tests of this repository are basically the same as this example, except
 the tests run on the three current firmwares (NanoS, NanoX and NanoS+).
+
+## Documentation
+
+You'll need the `doc` dependencies to generate the documentation:
```
fbeutin-ledger, line 133: The complete documentation can be found `[here](https://ledgerhq.github.io/ragger/).`
If you want to generate it locally, you'll need the `doc` ...

doc/faq.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Why all my tests are raising a ``ConnectionError`` when using SpeculosBackend?
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+This indicates that :term:`Speculos` could not reach its server, generally
+because the emulator could not be started.
+
+With verbose options, you should be able to find in the logs the Speculos
+command line, which should look like:
+
```
fbeutin-ledger, line 19: + check if a speculos process is already started

doc/tutorial.rst
```diff
@@ -0,0 +1,388 @@
...
+      with backend:
+          yield backend
+
+.. code-block:: python
+  :caption: test_something.py
+  :linenos:
+
+  def test_something(backend):
+      backend.exchange(cla=0xE0, ins=0x01, data=bytes.fromhex("DEADBEEF"))
+      backend.right_click()
```
fbeutin-ledger, line 112: Dans tous les tests on renome ce backend `client`...

doc/tutorial.rst
```diff
@@ -0,0 +1,388 @@
...
+      backend = SpeculosBackend(app_location, firmware)
+      with backend:
+          yield backend
+
+And now we can easily specify which firmware should be used:
+
+.. code-block:: bash
+
+  $ # start the application on a physical device connected to the computer, or on an emulator
+  $ pytest -v --nanosp --backend ledgercomm test_something.py
```
fbeutin-ledger, line 381: Il faut que le test_something prenne en paramètre le firmware pour qu'il soit fixturé non ? Peut être donner le nouveau template de test_something

fbeutin-ledger (approved):

.github/workflows/documentation.yml
lpascal-ledger: Because currently we deploy from `develop`, and so I though it was better for the doc to be synced with develop.
Originally, I was keeping `master` for release available on `pypi.org`, not `test.pypi.org`. But this can change.

doc/tutorial.rst
lpascal-ledger: Yes, not sure it is a good thing we call it "client", as it may be confusing with the actual application client, which is to the dev to develop. As this documentation makes clear difference between the two, I preferred using the "right" name.

README.md
lpascal-ledger: Yes, + I forgot to document the fact that `graphviz` is needed.

doc/faq.rst
lpascal-ledger: Indeed!

doc/tutorial.rst
lpascal-ledger: Hmm as the `backend` fixture is itself fixtured with `firmware`, this is enough no? Lemme check.

lpascal-ledger: Yep that's cool!

.github/workflows/documentation.yml
fbeutin-ledger: No problem with either `master` or `develop`, but I think the workflow should only update the doc on `push on develop` otherwise we will update with draft content

doc/tutorial.rst
fbeutin-ledger: Fine by me, how about making a go on all existing apps, changing `client` by `backend`, and removing the current `backend` that serves no purpose ?
We should do it now before it's out of hand

.github/workflows/documentation.yml
lpascal-ledger: It is already the case as the workflow triggers only on `push` master & develop (other are PR), but I agree it is not very clear. I'll update the condition.

doc/tutorial.rst
lpascal-ledger: Current `backend` as a purpose as an option, could be renamed `backend_choice` or `backend_option` or whatever.

I'll update the template provided by Ragger, and I agree we should update other apps accordingly now.

fbeutin-ledger (approved):

PR #43, opened 2022-11-15
Add dispatchable readme to go alongside the conftest

fbeutin-ledger:


fbeutin-ledger:
I've created a `template` directory dedicated to hosting the conftest and usage file

xchapron-ledger:
Can you also update the conftest template link in the readme (https://github.com/LedgerHQ/ragger/blob/develop/README.md?plain=1#L81)

template/usage.md
```diff
@@ -0,0 +1,55 @@
...
+```
+pip install --extra-index-url https://test.pypi.org/simple/ -r requirements.txt
+sudo apt-get update && sudo apt-get install qemu-user-static
+```
+
+### Compile the application
+
+For all required devices, compile the application by running the following command in the container `ghcr.io/ledgerhq/ledger-app-builder/ledger-app-builder-lite:latest`
+```
+make BOLOS_SDK=$<device>_SDK
```
xchapron-ledger, line 19: You added more details here: https://github.com/LedgerHQ/app-elrond/pull/19/files
Do you think these should be added on every repo?
Might not be as some are more related to building and sideloading more than to testing. Hence theses instructions might more be located at the root of the repo instead of in the test folder?

PR #52, opened 2022-11-22
backend: Sanitize screen and snapshot comparison

### LINKED ISSUE #51: Rethink the tmp screen saving
Currently ragger speculos :
- takes a screenshot of the screen and saves it in tmp
- takes a screenshot of the screen and compares it with the golden

This is to align the speculos behavior with the ledgerwallet behavior `compare_screen_with_snapshot`

I had a timing issue where the tmp screenshot was different from the screenshot taken for the comparison, we should only take one screenshot

xchapron-ledger:
Before this commit, there was 3 screenshot taken over Speculos when comparing screen with golden_run:
- one for temp folder saving
- one for golden folder saving
- one for comparison with golden folder

Which could leads in hard to understand issues if the screen changes between one of theses events.

Therefore, this commit introduce a patch which allow to take only one screenshot and uses it for:
- temp folder saving
- golden folder saving if golden_run
- screen comparison with golden snap

all of this with keeping somehow the same API which is "compatible" with ledgercomm and ledgerwallet as it can be implemented as:
```
def compare_screen_with_snapshot(self,
                                 golden_snap_path: Path,
                                 crop: Optional[Crop] = None,
                                 tmp_snap_path: Optional[Path] = None,
                                 golden_run: bool = False) -> bool:

    # Ignore tmp_snap_path which is not possible on this backend

    if golden_run:
        return True

    # TODO:
    # Ask the user to compare the screen with the golden_snap_path
    # while taking into account the optional crop parameter
    return True
```

Fixes #51

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,6 @@ The `src/ragger/backend/interface.py` file describes the methods that can be imp
 * `left_click`: perform a left click on a device.
 * `both_click`: perform a click on both buttons (left + right) of a device.
 * `compare_screen_with_snapshot`: compare the current device screen with the provided snapshot.
-* `save_screen_snapshot`: save the current device screen as a snapshot.
 
 The `src/ragger/navigator/navigator.py` file describes the methods that can be implemented by the different device navigators and that allow to interact with an emulated device:
 * `navigate`: navigate on the device according to a set of navigation instructions provided.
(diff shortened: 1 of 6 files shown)


fbeutin-ledger (approved):
+ bump la version, on peut considérer que c'est un fix ?

PR #55, opened 2022-11-29
Fix usage of backend instead of client in pytest

fbeutin-ledger:


template/conftest.py
```diff
@@ -7,8 +7,10 @@
 
 
 # This variable is needed for Speculos only (physical tests need the application to be already installed)
+# Adapt this path to your 'tests/elfs' directory
 APPS_DIRECTORY = (Path(__file__).parent.parent / "elfs").resolve()
 
+# Adapt this path to the APPNAME in your Makefile
```
lpascal-ledger, line 13: Not really the APPNAME of the Makefile, but rather the `<name>_<device>.elf` used to store ELF binaries.

lpascal-ledger (approved):

lpascal-ledger (approved):

lpascal-ledger:
As a template bug fix, this should be deployed ASAP, that's a good candidate for a `0.9.1`

xchapron-ledger (commented):
I think we should update `ragger/tests/conftest.py` in this Pr too?

fbeutin-ledger:
> I think we should update `ragger/tests/conftest.py` in this Pr too?

I suggest updating it next time we add tests in ragger, otherwise it's a bit of work for no gain

xchapron-ledger (approved):

PR #59, opened 2022-12-16
Add typed marker

fbeutin-ledger:


fbeutin-ledger:
There is no py.typed marker so this generates errors during typing checks for modules using ragger

PR #66, opened 2022-12-21
Add log_apdu_file option to ragger

fbeutin-ledger:


src/ragger/backend/ledgercomm.py
```diff
@@ -71,6 +73,7 @@ def __exit__(self, *args, **kwargs):
 
     def send_raw(self, data: bytes = b"") -> None:
         logger.debug("Sending '%s'", data)
+        self.apdu_logger.debug("=> %s", data.hex())
```
lpascal-ledger, line 76: Rather than multiplying the number of loggers, you could define several handler for `apdu_logger`, one on the `stdout` (which would then take place of the previous `logger`) and the second, optional, which would log into a file.

fbeutin-ledger: Done

src/ragger/backend/interface.py
```diff
@@ -42,6 +44,24 @@ def __init__(self, firmware: Firmware):
...
+            apdu_handler = logging.FileHandler(filename=log_apdu_file, mode='w', delay=True)
+            apdu_handler.setFormatter(logging.Formatter('%(message)s'))
+            self.apdu_logger.addHandler(apdu_handler)
+            atexit.register(self._cleanup)
+
+    def _cleanup(self):
+        for handler in self.apdu_logger.handlers:
+            if isinstance(handler, logging.FileHandler):
+                handler.close()
+
```
lpascal-ledger, line 64: I feel like all this configuration should be in a specific module, rather than buried here (in particular the log path as an argument feels strange to me). There can be a `ragger.logger` module with an `init` function, which would initialize every needed loggers (this one but also the `root` one currently in `ragger.__init__.py`).
These loggers could be initialized explicitly, or implicitly in `ragger.__init__.py` for instance.

```python
import atexit
import logging 
from pathlib import Path
from typing import Optional

def init(log_file: Path, format: Optional[str] = None):
    format = format or '[%(asctime)s][%(levelname)s] %(name)s - %(message)s'

    logger = logging.getLogger(__package__)
    logger.setLevel(level=logging.DEBUG)
    handler = logging.StreamHandler()
    handler.setFormatter(logging.Formatter(format))
    logger.addHandler(handler)

    apdu_logger = logging.getLogger("apdu_logger")
    apdu_logger.setLevel(level=logging.DEBUG)
    handler = logging.StreamHandler()
    handler.setFormatter(logging.Formatter(format))
    apdu_logger.addHandler(handler)

    if log_file is not None:
        apdu_handler = logging.FileHandler(filename=log_file, mode='w', delay=True)
        apdu_handler.setFormatter(logging.Formatter('%(message)s'))
        apdu_logger.addHandler(apdu_handler)
        
       def cleanup():
            for handler in apdu_logger.handlers:
                if isinstance(handler, logging.FileHandler):
                    handler.close()

        atexit.register(cleanup)
```

```python
from ragger import loggers

loggers.init(<args>)
```

Doing so, everybody could now use the loggers everywhere with a `logging.getLogger(<name>)`. That would be cleaner that the current behavior, where `logger` is a global variable fetched with `from ragger import logger`.

tests/unit/backend/test_interface.py
```diff
@@ -101,3 +101,17 @@ def test_exchange_async(self):
             pass
         self.assertTrue(self.backend.mock.exchange_async_raw.called)
         self.assertEqual(self.backend.mock.exchange_async_raw.call_args, ((expected, ), ))
+
+
+class TestBackendInterfaceLogging(TestCase):
+
+    def test_log_apdu(self):
+        self.firmware = _Firmware("nanos", "2.0.1", "other")
+        test_file = Path("/tmp/test_log_file.log").resolve()
```
lpascal-ledger, line 110: If you want to clean up all created file / directory after you test is done, you can use the `tempfile` buildin library:

```python
with tempfile.TemporaryDirectory() as td:
    test_file = Path(td.name) / "test_log_file.log"
   [...]
```

The context manager will automatically remove the directory after the tests.

lpascal-ledger (approved):

xchapron-ledger:
@fbeutin-ledger can you rebase it and add your changes in the changelog?

fbeutin-ledger:
> @fbeutin-ledger can you rebase it and add your changes in the changelog?

Done

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - navigator: Improve navigation by waiting screen changes instead of hardcoded sleeps
 - navigator: navigate_and_compare(): Add optional snap_start_idx parameter
 - navigator: navigate_until_text(): take a list of validation instructions and check last screen
+- conftest: `--log_apdu_file <filename>` option
```
lpascal-ledger, line 23: 1.1.0 is already deployed no? This should be in the `## [Unreleased]` entry.

lpascal-ledger (changes requested):

CHANGELOG.md
fbeutin-ledger: Done

PR #72, opened 2022-12-27
Navigation improvments

xchapron-ledger:
This PR contains:
- A first commit replacing navigation sleeps by waits for screen changes and using an backend internal cache for the context so that there is no need to initialize the cache at the start of the navigate call.
- A second commit allowing to have snapshot idx not starting at 0 for navigate_and_compare().
- A final commit that change the navigate_until_text behavior so that it:
  - takes a list of validation instructions (needed for Ledger Stax (approve + wait status message)
  - checks the screen after the last validation instruction, wasn't done before.

There is still one corner case which can brings issues with the wait_for_screen_change mechanism. It's when some instructions are performs without using the navigator, and then other are done with the navigator.
In this cases, the backend internal context is not up to date. In theses cases it can be updated with a call to navigator.get_current_screen_content(). 

Usage of this features can be seen:
- in this PR commit for nano usage: https://github.com/LedgerHQ/stax-eos/pull/2/commits/976f40e7722dc7bcca99e745c9a8762540871937
- in this PR commit for stax usage: https://github.com/LedgerHQ/stax-eos/pull/2/commits/b6fa0a7926805d218590fb66fcd7d34933fc23b2


src/ragger/backend/speculos.py
```diff
@@ -74,6 +74,7 @@ def __init__(self,
                                                       api_url=self.url,
                                                       **kwargs)
         self._pending: Optional[ApduResponse] = None
+        self._screen_content: dict = {}
```
agrojean-ledger, line 77: @xchapron-ledger so now we require the context to be some kind of dict ? (and not "Any")

agrojean-ledger: Oh I see, since it's the backend and not the navigator we can assume it will be a dict. Disregard my comment.

xchapron-ledger: It's for something not exposed.
On the `interface` it's still declared as `Any`, but in the internal of `speculos` backend we known that in this case it is a `Dict`.

src/ragger/navigator/navigator.py
```diff
@@ -394,51 +406,49 @@ def navigate_until_text_and_compare(self,
...
     def navigate_until_text(self,
                             navigate_instruction: NavIns,
-                            validation_instruction: NavIns,
+                            validation_instructions: List[NavIns],
                             text: str,
-                            timeout: int = 30) -> None:
+                            timeout: int = 30,
+                            screen_change_before_first_instruction: bool = True,
+                            screen_change_after_last_instruction: bool = True,
+                            snap_start_idx: int = 0) -> None:
```
agrojean-ledger, line 451: There should not be a `snap_start_idx` in the `navigate_until_text` arguments @xchapron-ledger

xchapron-ledger: Removed!

src/ragger/navigator/navigator.py
```diff
@@ -394,51 +406,48 @@ def navigate_until_text_and_compare(self,
...
-            if (time() - start > timeout):
+            remaining = timeout - (time() - start)
+            if (remaining < 0):
                 raise TimeoutError(f"Timeout waiting for text {text}")
 
-            ctx = self._backend.wait_for_screen_change(2.0, ctx)
+            self._backend.wait_for_screen_change(remaining)
+
+        # Perform navigation validation instructions in an "navigate_and_compare" way.
+        if validation_instructions:
```
agrojean-ledger, line 433: Do you need this check ? `validation_instructions` is a mandatory parameter.

xchapron-ledger: It could be an empty list. In the case where it's called from `navigate_until_text` but there is nothing that should be done once the screen is found.

agrojean-ledger: True.

src/ragger/backend/speculos.py
```diff
@@ -159,22 +171,25 @@ def compare_screen_with_snapshot(self,
...
-                if (time() - start > timeout):
-                    raise TimeoutError("Timeout waiting for screen change")
-                content = self._client.get_current_screen_content()
-            return content
-
-    def compare_screen_with_text(self, text: str):
-        return text in dumps(self._client.get_current_screen_content())
-
-    def get_current_screen_content(self) -> Any:
+    def _get_current_screen_content(self) -> dict:
```
agrojean-ledger, line 174: I find having  `_get_current_screen_content` and  `get_current_screen_content` a bit confusing @xchapron-ledger

xchapron-ledger: @agrojean-ledger I've remove the dummy  `_get_current_screen_content` "helper".

agrojean-ledger (dismissed):

PR #75, opened 2023-01-03
Minor navigator tweaks

lpascal-ledger:
- adding a `Navigator:add_callback` method (allows to use a predefined Navigator such as `FatstacksNavigator` with adding/replacing some callbacks with custom ones)
- allowing `Navigator` methods to use `NavInsID` as well as `NavIns` as instruction: lots of callbacks are called without arguments (especially when using `UseCase*`), so it makes sense to call these methods with `[NavInsID.A, NavInsID.B, NavIns(NavInsID.C, (4, ))]` instead of `[NavIns(NavInsID.A), NavIns(NavInsID.B), NavIns(NavInsID.C, (4, ))]`
- `Navigator:navigate_and_compare` method: `screen_change_before_first_instruction` and `screen_change_after_last_instruction` set to `False` by default: default behavior of the method is to execute the given instructions, whatever the context. On the other way around, no instruction is executed, and the method raises a `TimeoutError`, which is an unclear behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - firmware: `Firmware:have_bagl` and `Firmware:have_nbgl` properties indicates which graphical
             interface the firmware implements.
 - navigator: Add USE_CASE_ADDRESS_CONFIRMATION_TAP
+- navigator: `add_callback()` method to register callbacks after Navigator initialization
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - navigator: Add USE_CASE_ADDRESS_CONFIRMATION_TAP
 - navigator: `add_callback()` method to register callbacks after Navigator initialization
 
+### Changed
+
+- navigator: `NavIns` arguments of `navigate()` and higher-level methods can now also be `NavInsID`,
+             which will automatically be converted to `NavIns` (with no argument for the callback)
+
 ### Fixed
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Fatstacks UseCaseHomeExt
+- Fatstacks `UseCaseHomeExt`
 - navigator: Improve navigation by waiting screen changes instead of hardcoded sleeps
-- navigator: navigate_and_compare(): Add optional snap_start_idx parameter
-- navigator: navigate_until_text(): take a list of validation instructions and check last screen
+- navigator: `navigate_and_compare()`: add optional snap_start_idx parameter
+- navigator: `navigate_until_text()`: take a list of validation instructions and check last screen
 
 
 ## [1.0.0] - 2022-12-26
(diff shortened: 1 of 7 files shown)


xchapron-ledger:
> Navigator:navigate_and_compare method: screen_change_before_first_instruction and screen_change_after_last_instruction set to False by default: default behavior of the method is to execute the given instructions, whatever the context. On the other way around, no instruction is executed, and the method raises a TimeoutError, which is an unclear behavior.

@lpascal-ledger I disagree on this one. navigate_and_compare is mostly used for navigation after APDU sending for which both before_first and after_last should de set to True.and therefore by default

xchapron-ledger:
> allowing Navigator methods to use NavInsID as well as NavIns as instruction

This will make the API a bit more confusing, but this is definitively saving a lot of useless `NavIns()` :fire:

> adding a Navigator:add_callback method

Just being curious here, for what use case do you need this?

lpascal-ledger:
I believe it is not so much more confusing, as the current behavior stays the same, and it only impacts those bored enough to read the doc and figure out they can do without specifying the `NavIns` when there is no arguments. But if it is too confusing, we could think of a way to make it more straightforward in term of code.

The use case for `add_callback` is any customization of an existing `Navigator`. When testing the `passwords` application, I started with using the `FatstacksNavigator`. This navigator holds lots of action but not all I had. So I added some custom, higher level functions which I feel where more clear. For instance, I have a menu with 5 choices, so instead of having instructions such as `NavIns(NavInsID.CHOICE_CHOOSE, (1, ))` or `NavIns(NavInsID.CHOICE_CHOOSE, (2, ))`, I created instructions `NavInsID.MENU_TO_CREATE` and `NavInsID.MENU_TO_CREATE`. For this I just initialize the navigator then add (sometimes override the pre-registered callbacks with) mine.
It is not that simple actually, as the NavInsID enum is not extendable - which is a problem in itself I've not found an easy fix. But one can declare another enum starting a `FIRST = max(i.value for i in NavInsID)`. It's just that the typing will creak a bit.

lpascal-ledger:
Regarding the default parameter change
> navigate_and_compare is mostly used for navigation after APDU sending for which both before_first and after_last should be set to True.
> The current value is True, and changing it is a breaking change that we can avoid.

I don't think these are better arguments than clarity of the method. `navigate_and_compare` should by default "navigate and compare" (that's what the name and the docstring suggest), not "wait for some stuff, then navigate and compare, then wait again for some other stuff". I agree the current usage is a major one, and so we should have `navigate_and_compare_after_screen_change` (or even `navigate_and_compare_after_trigger` with a customizable callback argument so that a user could wait for a specific screenshot or RAPDU for instance, no just the first screen change). We already have methods with long and explicit names, so have `navigate_and_compare` taken by a method doing more that just that feels wrong.

> navigate_until_text and navigate_until_text_and_compare should behave similarly and they are currently also set with True as default.

This is true indeed.

> If we put the default value to False, by default no wait is done, and it immediately leads to racing issues, which is worse than an obscur TimeoutError that will occur in rare situations.

Racing issue? I thought the error would rather be a snapshot comparison failure which to my mind is easier to understand ("oh right of course it's not the welcome screen, I should wait for the APDU to trigger the transaction") than a timeout error, which needs to dive into the code to understand why it does not work as expected. Indeed a race would not be good, but it would hopefully be avoided by using a function with an explicit name.

xchapron-ledger:
> Racing issue? I thought the error would rather be a snapshot comparison failure which to my mind is easier to understand ("oh right of course it's not the welcome screen, I should wait for the APDU to trigger the transaction") than a timeout error, which needs to dive into the code to understand why it does not work as expected. Indeed a race would not be good, but it would hopefully be avoided by using a function with an explicit name.

Depending on the speed of processing of `speculos` vs the speed of execution of the `navigator`, either the welcome screen or the first review screen will be shown. This is definitively a racing issue that will depend on the computer running the test and the different threads it's running in the same time, so somehow something a bit nasty.

xchapron-ledger (approved):

PR #76, opened 2023-01-04
Poc of interface guiding user to physical device instructions

lpascal-ledger:
Testable with boilerplate application, for instance (on nanos):

```
pytest tests/test_pubkey_cmd.py -k get_public_key_confirm --device nanos --backend ledgercomm --tb=short
```


src/ragger/navigator/__init__.py
```diff
@@ -13,7 +13,8 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 """
-from .navigator import NavInsID, NavIns, Navigator
+from .instruction import NavInsID, NavIns
+from .navigator import Navigator
```
xchapron-ledger, line 17: This is a breaking change that might be mentioned somewhere in the changelog?

xchapron-ledger (commented):
I didn't review the gui code but that looks promising, I need to take some time testing this!

src/ragger/navigator/__init__.py
lpascal-ledger: This should not break anything, these variables should still be available on every paths they were available previously. These imports are still valid:
- `from ragger.navigator import NavInsID, NavIns, Navigator`
- `from ragger.navigator.navigator import NavInsID, NavIns, Navigator`

xchapron-ledger: Great then!

src/ragger/conftest/base_conftest.py
```diff
@@ -153,9 +153,12 @@ def prepare_speculos_args(root_pytest_dir: Path, firmware: Firmware, display: bo
 def create_backend(root_pytest_dir: Path, backend_name: str, firmware: Firmware, display: bool,
                    log_apdu_file: Optional[Path], cli_user_seed: str):
     if backend_name.lower() == "ledgercomm":
-        return LedgerCommBackend(firmware=firmware, interface="hid", log_apdu_file=log_apdu_file)
+        return LedgerCommBackend(firmware=firmware,
+                                 interface="hid",
+                                 log_apdu_file=log_apdu_file,
+                                 with_gui=True)
```
xchapron-ledger, line 159:
```suggestion
                                 with_gui=display)
```
What do you think of reusing this parameter?

lpascal-ledger: Makes sense. I'll also add some doc to explain this, it could be confusing in a way.

xchapron-ledger (dismissed):

PR #77, opened 2023-01-04
navigator: Add USE_CASE_ADDRESS_CONFIRMATION_TAP

xchapron-ledger:
Needed for situation where nbgl_useCaseAddressConfirmationExt is used.


src/ragger/firmware/fatstacks/positions.py
```diff
@@ -233,6 +233,7 @@ def __iter__(self):
             "next": BUTTON_LOWER_RIGHT,
         },
         "UseCaseAddressConfirmation": {
+            "tap": BUTTON_ABOVE_LOWER_MIDDLE,
```
agrojean-ledger, line 236: so what's the difference with 'confirm' ?

xchapron-ledger: There is no difference except that the "confirm" action is meant for the "Confirm" button, and the "tap" action is meant for the "Tap to continue" action.
I think that having both help understand what is supposed to be done when they are in an navigate_and_compare instruction list for example.

agrojean-ledger: Ok, I see. It is less confusing with a dedicated instruction indeed.

agrojean-ledger:
@xchapron-ledger no test required for this change ?

src/ragger/firmware/fatstacks/positions.py
lpascal-ledger: So in order to keep meaningful names, could it be "continue" or is it too restrictive?

xchapron-ledger:
> @xchapron-ledger no test required for this change ?

We can't do functional test as speculos for stax is not public yet. and there is no unit-test for the use_cases. I don't think we can do much?
Maybe just use them one by one and check that the touch function is called with the right positions? I'm not sure this will help?

src/ragger/firmware/fatstacks/positions.py
xchapron-ledger: I've named it "Tap" to be equivalent with the already existing `USE_CASE_REVIEW_TAP`

lpascal-ledger:
Declarative code such as those are not unit-testable: it's just syntax, so a test would just check if "a" is affected into "b".
Functional test would be interesting, be very heavy: we should build an application mimicking all of Fatstacks screens and run tests on it.
As it is a bit ambitious for not that much gain, maybe when released we could integrate the `boilerplate` functional test suite into the `ragger` CI, and that would be good enough (?)

lpascal-ledger (approved):

PR #81, opened 2023-01-05
Conftests improvments

xchapron-ledger:


CHANGELOG.md
```diff
@@ -8,8 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- conftest: `--elfs_dir` argument to specify elfs directories
+- conftest: `--all` argument to run tests on all devices
 
 ### Changed
+- conftest: Default to using elf file in `bin/app.elf` if `--elfs_dir` is not specified
+- conftest: Targeted device must be specified, in doesn't default to all by default anymore
```
fbeutin-ledger, line 16: So by default it will run on `./bin/app.elf` ? So if pytest is ran from the project root or the test root we don't have the same behavior?

If we specify a directory we will look for `<dir>/<appname>_<device>.elf`
I'll be honest I find this really inconsistent and a regression to what we currently have :/

template/conftest.py
```diff
@@ -68,53 +74,66 @@ def pytest_generate_tests(metafunc):
...
+def prepare_speculos_args(firmware: Firmware, display: bool, elfs_dir: str):
     speculos_args = []
 
     if display:
         speculos_args += ["--display", "qt"]
 
-    app_path = app_path_from_app_name(APPS_DIRECTORY, APP_NAME, firmware.device)
+    if elfs_dir == "bin":
+        app_path = APP_ROOT_DIR / elfs_dir / "app.elf"
+        assert app_path.is_file(), f"{app_path} must exist"
```
fbeutin-ledger, line 105: running a basic `pytest` with no argument in project root will trigger a weird 
`app-X/tests/bin must exist` error?

template/conftest.py
```diff
@@ -68,53 +74,66 @@ def pytest_generate_tests(metafunc):
...
+    else:
+        app_path = app_path_from_app_name(APP_ROOT_DIR / elfs_dir, APP_NAME, firmware.device)
```
fbeutin-ledger, line 107: Very different behavior from the default one

template/conftest.py
```diff
@@ -68,53 +74,66 @@ def pytest_generate_tests(metafunc):
...
 
 
-# This final fixture will return the properly configured backend, to be used in tests
-@pytest.fixture
-def backend(backend_name, firmware, display, log_apdu_file):
-    with create_backend(backend_name, firmware, display, log_apdu_file) as b:
+# This fixture will return the properly configured backend, to be used in tests.
+# If your tests needs to be run on independent Speculos instances (in case they affect
+# setting for example), then you should change this fixture scope.
+@pytest.fixture(scope="session")
```
fbeutin-ledger, line 130: Can't this already be done by using test classes that instantiate the backend as a class fixture?
If it is the case we should document it and try it on some of the boilerplate tests, this would allow a finetuning of the backend scope

Otherwise it would be nice to be able to have multiple backend scopes: per file / per test / for everyone / etc

template/conftest.py
```diff
@@ -68,53 +74,66 @@ def pytest_generate_tests(metafunc):
...
+def backend(backend_name, firmware, display, elfs_dir, log_apdu_file):
+    with create_backend(backend_name, firmware, display, elfs_dir, log_apdu_file) as b:
         yield b
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
```
fbeutin-ledger, line 136: No reason not to have the navigator a session scope, it will improve test speed 👍

fbeutin-ledger (approved):

PR #82, opened 2023-01-06
template: conftest.py: Use session scope for firmware and backend fixtures

xchapron-ledger:


template/conftest.py
```diff
@@ -107,8 +107,12 @@ def create_backend(backend_name: str, firmware: Firmware, display: bool, log_apd
...
 
 
-# This final fixture will return the properly configured backend, to be used in tests
-@pytest.fixture
+# This fixture will return the properly configured backend, to be used in tests.
+# As Speculos instantiation takes some time, this fixture scope is by default "session".
+# If your tests needs to be run on independent Speculos instances (in case they affect
+# settings for example), then you should change this fixture scope and choose between
+# function, class, module or session.
+@pytest.fixture(scope="session")
```
fbeutin-ledger, line 115: What happens when the user runs tests on all devices ?

xchapron-ledger: Then tests are run device per device (e.g. all tests on device1, then all tests on device2), and the backend is only instantiated once per device.
So perfectly working but not obvious why :)

lpascal-ledger (approved):

PR #83, opened 2023-01-06
template: conftest.py: Targeted device must be specified with new --device required option

xchapron-ledger:


template/conftest.py
```diff
@@ -16,19 +16,19 @@
 
 BACKENDS = ["speculos", "ledgercomm", "ledgerwallet"]
 
+DEVICES = ["nanos", "nanox", "nanosp", "all"]
```
fbeutin-ledger, line 19: + stax as well

template/conftest.py
```diff
@@ -16,19 +16,19 @@
...
+
 FIRMWARES = [Firmware('nanos', '2.1'),
              Firmware('nanox', '2.0.2'),
              Firmware('nanosp', '1.0.3')]
 
 
 def pytest_addoption(parser):
-    parser.addoption("--backend", action="store", default="speculos")
+    parser.addoption("--device", choices=DEVICES, required=True)
```
fbeutin-ledger, line 27: default="all" would be nice IMO

template/usage.md
```diff
@@ -73,12 +73,10 @@ Standard useful pytest options
 
 Custom pytest options
 ```
+    --device <device>           run the test on the specified device [nanos,nanox,nanosp,all]. This parameter is mandatory
```
fbeutin-ledger, line 76: +stax

template/conftest.py
xchapron-ledger: I think we should add when SDK + speculos is public and Stax code name is fixed.

template/conftest.py
xchapron-ledger: I rather have default=error so that new users easily understand they should specify the device :)
And also maybe so that later we can have default=detected last build?

template/usage.md
xchapron-ledger: see https://github.com/LedgerHQ/ragger/pull/83#discussion_r1063448791

lpascal-ledger (approved):

PR #85, opened 2023-01-09
Add a check_dependencies_compatibility Ci step

fbeutin-ledger:


.github/workflows/build_and_tests.yml
```diff
@@ -21,7 +27,7 @@ jobs:
       run: sudo apt-get update && sudo apt-get install -y qemu-user-static
     - name: Build & install
       run: |
-        pip install --extra-index-url https://test.pypi.org/simple/ -U .[tests,all_backends]
+        pip install --extra-index-url https://test.pypi.org/simple/ -U .[tests,speculos]
```
lpascal-ledger, line 30: So `PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python` should not be superfluous! :)

.github/workflows/build_and_tests.yml
```diff
@@ -11,6 +11,12 @@ on:
       - master
 
 jobs:
+  check_dependencies_compatibility:
+    runs-on: ubuntu-latest
+    - name: Check dependencies compatibility
+      run: |
+        pip install --extra-index-url https://test.pypi.org/simple/ -U .[tests,all_backends]
```
lpascal-ledger, line 18: You can add a `python -c "from ragger.backends import *` just to trigger the dependencies high-level imports and check all is nice.

fbeutin-ledger: I've added the check for all top level modules

lpascal-ledger (approved):

PR #90, opened 2023-01-18
[fix] Suggestion button positions

lpascal-ledger:


xchapron-ledger (approved):
This is a BOLOS change from what was exposed in a previous rc I guess?

lpascal-ledger:
@xchapron-ledger no, as a layout BOLOS allows the suggestion buttons to be displayed wherever we want.
**However**, the BOLOS screens integrating suggestion buttons (the passphrase onboarding) place them _under_ the text area. When I wrote this portion of `ragger`, I was testing with my version of `Recovery Check`, where suggestion buttons are _above_ the text area. I corrected this for UI consistency, and figured that by default they ought to be placed there.

PR #95, opened 2023-01-25
Use a single common conftest with configuration

fbeutin-ledger:


src/ragger/conftest/base_conftest.py
```diff
@@ -0,0 +1,171 @@
...
+BACKENDS = ["speculos", "ledgercomm", "ledgerwallet"]
+
+DEVICES = ["nanos", "nanox", "nanosp", "all"]
+
+FIRMWARES = [Firmware('nanos', '2.1'),
+             Firmware('nanox', '2.0.2'),
+             Firmware('nanosp', '1.0.3')]
+
+
+def pytest_addoption(parser):
```
xchapron-ledger, line 20: On app-u2f, I've added more options: https://github.com/LedgerHQ/app-u2f/blob/develop/tests/speculos/conftest.py#L30-L31
How can it be handled?

xchapron-ledger: I also have a client fixture, but this could be added on the app conftest

fbeutin-ledger: As agreed on slack, let's keep them in mind for a v2

src/ragger/conftest/base_conftest.py
```diff
@@ -0,0 +1,170 @@
...
+    missing = []
+    for key, value in configuration.REQUIRED_CONFIGURATION.items():
+        if value == "":
+            missing.append(key)
+    if missing:
+        raise ValueError(f"Missing required configuration parameters: '{missing}")
+
+
+# RUN ON IMPORT
+print_full_conf()
```
xchapron-ledger, line 169: This is a bit verbose? Or not that much?

src/ragger/conftest/configuration.py
```diff
@@ -0,0 +1,30 @@
+from typing import Dict, Any
+
+# The following parameters are required.
+# Please set their values in your 'conftest.py' file
+REQUIRED_CONFIGURATION: Dict[str, Any] = {
+
+    # Use the real name of your application
+    # example: "boilerplate"
```
xchapron-ledger, line 8: Maybe something like: It is used to compute the elf name with template `<APP_NAME>_<DEVICE_NAME>.elf`

src/ragger/conftest/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+    # Adapt this path to your 'tests/elfs' directory
+    # example: Path(__file__).parent.parent / "elfs").resolve()
+    "APPS_DIRECTORY": ""
+}
+
+# The following parameters are optionnal and can be used to customize your tests
+# You can set their values in your 'conftest.py' file if you want to use them
+OPTIONNAL_CONFIGURATION: Dict[str, Any] = {
+
+    # use this parameter if you want Speculos to emulate other applications being installed on the device
```
xchapron-ledger, line 21: Again something about location from where elf are pick?

src/ragger/conftest/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+    # example: {"bitcoin": "Bitcoin", "ethereum": "Ethereum"}
+    "SIDELOADED_APPS": {},
+
+    # As the backend instantiation may take some time, Ragger supports multiple backend scopes.
+    # You can choose to share the backend instance between {session / module / class / function}
+    # When using "session" all your tests will share a single backend instance (faster)
+    # When using "function"each test will have its independant backend instance (more precise)
```
xchapron-ledger, line 28:
```suggestion
    # When using "function" each test will have its independant backend instance (no collusion)
```
?

template/conftest.py
```diff
@@ -1,149 +1,24 @@
...
+# Change this path to the path in which your compiled elf files are located
+configuration.REQUIRED_CONFIGURATION["APPS_DIRECTORY"] = (Path(__file__).parent / "elfs").resolve()
 
-@pytest.fixture
-def use_only_on_backend(request, backend_name):
-    if request.node.get_closest_marker('use_on_backend'):
-        current_backend = request.node.get_closest_marker('use_on_backend').args[0]
-        if current_backend != backend_name:
-            pytest.skip(f'skipped on this backend: "{current_backend}"')
+# You can configure optional parameters by overiding the value of ragger.conftest.OPTIONNAL_CONFIGURATION
```
xchapron-ledger, line 17: Can you add in comment version here with their default value?

xchapron-ledger (approved):

template/conftest.py
```diff
@@ -1,149 +1,24 @@
...
-
-
-@pytest.fixture
-def navigator(backend, firmware, golden_run):
-    if firmware.device.startswith("nano"):
-        return NanoNavigator(backend, firmware, golden_run)
-    else:
-        raise ValueError(f"Device '{firmware.device}' is unsupported.")
+# Change this value to the name of your application
+configuration.REQUIRED_CONFIGURATION["APP_NAME"] = "my_app"
```
lpascal-ledger, line 12: Don't you need to `from ragger.conftest import configuration` before using it?

src/ragger/conftest/base_conftest.py
lpascal-ledger: I agree, it is a bit much. Maybe using a proper `logger.debug` instead of `print` would avoid flooding the stdout all the time.

template/conftest.py
```diff
@@ -1,149 +1,24 @@
...
+### CONFIGURATION END ###
+#########################
 
-def pytest_configure(config):
-    config.addinivalue_line(
-        "markers",
-        "use_only_on_backend(backend): skip test if not on the specified backend",
-    )
+# Pull all features from the base ragger conftest using the overridden configuration
+from ragger.conftest.base_conftest import *
```
lpascal-ledger, line 24: This import is a bit dirty. Maybe it could work with the [pytest plugin mechanism](https://pytest.org/en/7.2.x/how-to/plugins.html#requiring-loading-plugins-in-a-test-module-or-conftest-file)?

```python
pytest_plugins = ("ragger.conftest.base_conftest",)
```

lpascal-ledger (commented):
All hail the misspell checker \o/

src/ragger/conftest/base_conftest.py
fbeutin-ledger: Done

template/conftest.py
fbeutin-ledger: Not convinced as we don't really want to have a dependency to update every time we change / remove / add in the configuration.
I added more text though to explain where to see the accepted values

src/ragger/conftest/configuration.py
fbeutin-ledger: Done

src/ragger/conftest/configuration.py
fbeutin-ledger: Done

template/conftest.py
fbeutin-ledger: Nice

.github/workflows/build_and_tests.yml
```diff
@@ -1,41 +1,54 @@
...
+      app_branch_name: master
+      app_binaries_artifact: boilerplate_binaries
+
   build_install_test:
-    name: Build, install and test the library
+    name: Install and test the library
+    needs: build_boilerplate_application
     runs-on: ubuntu-latest
     env:
       PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION: python
```
xchapron-ledger, line 26: Can probably be removed now?

src/ragger/conftest/base_conftest.py
```diff
@@ -0,0 +1,205 @@
...
+        device = "nanos2"
+
+    # Find the compiled application for the requested device
+    project_root_dir = find_project_root_dir(root_pytest_dir)
+
+    app_path = Path(project_root_dir / "build" / device / "bin" / "app.elf").resolve()
+    if not app_path.is_file():
+        raise ValueError(f"File '{app_path}' missing. Did you compile for this target?")
+
+    if not len(conf.OPTIONAL["SIDELOADED_APPS"]) == 0:
```
xchapron-ledger, line 114:
```suggestion
    if conf.OPTIONAL["SIDELOADED_APPS"]:
```
this is equivalent no?

src/ragger/conftest/configuration.py
```diff
@@ -0,0 +1,23 @@
...
+# The following parameters, if any, are required.
+# Please set their values in your 'conftest.py' file
+REQUIRED: Dict[str, Any] = {}
+
+# The following parameters are optional and can be used to customize your tests
+# You can set their values in your 'conftest.py' file if you want to use them
+OPTIONAL: Dict[str, Any] = {
+
+    # use this parameter if you want Speculos to emulate other applications being installed on the device
+    # example: {"bitcoin": "Bitcoin", "ethereum": "Ethereum"}
```
xchapron-ledger, line 12: Can you explain what should be use for the key and the value and what are both used for?

src/ragger/utils/misc.py
```diff
@@ -17,29 +17,17 @@
 from pathlib import Path
 
 
-def app_path_from_app_name(app_dir: Path, app_name: str, device: str) -> Path:
```
xchapron-ledger, line 20: Removing app_path_from_app_name means that this change is not retrocompatible.
Keeping it would make it an opt-in change (at conftest change). Maybe we can just deprecate it for now so that it doesn't break CI?
Or we can choose to explicitly break thing so that they are updated (and we don't have to keep deprecated stuffs).
What do you think about it?

template/conftest.py
```diff
@@ -1,156 +1,15 @@
...
-            pytest.skip(f'skipped on this backend: "{current_backend}"')
-
-
-def pytest_configure(config):
-    config.addinivalue_line(
-        "markers",
-        "use_only_on_backend(backend): skip test if not on the specified backend",
-    )
+# Pull all features from the base ragger conftest using the overridden configuration
+pytest_plugins = ("ragger.conftest.base_conftest", )
```
xchapron-ledger, line 15: can we add an easy link to use for user wanting to look at the conftest content without having to find the github repo, and then scroll in it? something like https://github.com/LedgerHQ/ragger/blob/develop/src/ragger/conftest.base_conftest.py

xchapron-ledger (approved):

src/ragger/conftest/base_conftest.py
```diff
@@ -0,0 +1,205 @@
+import pytest
+from typing import Optional
+from pathlib import Path
+from ragger.firmware import Firmware
+from ragger.backend import SpeculosBackend, LedgerCommBackend, LedgerWalletBackend
+from ragger.navigator import NanoNavigator, StaxNavigator
+from ragger.utils import find_project_root_dir
+from ragger.logger import get_default_logger
+
+from ragger.conftest import configuration as conf
```
lpascal-ledger, line 10: To avoid passing through `ragger/conftest/__init__.py` which could end up into circular import, I think relative import is better when inside the same module.
```suggestion
from . import configuration as conf
```

src/ragger/conftest/base_conftest.py
```diff
@@ -0,0 +1,205 @@
...
+        "markers",
+        "use_only_on_backend(backend): skip test if not on the specified backend",
+    )
+
+
+def log_full_conf():
+    logger = get_default_logger()
+    logger.debug("Running Ragger with the following conf:")
+    for key, value in conf.REQUIRED.items():
+        logger.debug(f"    {key} = '{value}'")
```
lpascal-ledger, line 189:
```suggestion
        logger.debug(f"\t{key} = '{value}'")
```

src/ragger/conftest/base_conftest.py
```diff
@@ -0,0 +1,205 @@
...
+    for key, value in conf.OPTIONAL.items():
+        logger.debug(f"    {key} = '{value}'")
```
lpascal-ledger, line 191:
```suggestion
        logger.debug(f"\t{key} = '{value}'")
```

src/ragger/conftest/configuration.py
```diff
@@ -0,0 +1,23 @@
+from typing import Dict, Any
+
+# The following parameters, if any, are required.
+# Please set their values in your 'conftest.py' file
+REQUIRED: Dict[str, Any] = {}
```
lpascal-ledger, line 5: As there is nothing `REQUIRED` for now, I'm in favor of removing all the related code. It's just noise.

src/ragger/conftest/configuration.py
```diff
@@ -0,0 +1,23 @@
...
+
+# The following parameters are optional and can be used to customize your tests
+# You can set their values in your 'conftest.py' file if you want to use them
+OPTIONAL: Dict[str, Any] = {
```
lpascal-ledger, line 9: I'm not keen on using a dictionary at this level. By construction, a dictionary does not hold specific values, just whatever keys, and attached values. Add to this that the dict is called `OPTIONAL`, and I could think as a dev as all these keys/values are totally optional and I could remove them.

I would be  wrong of course, as the conftest file accesses variables such as `conf.OPTIONAL["BACKEND_SCOPE"]` which would fail if `BACKEND_SCOPE` is not defined, or makes assumption on what kind of variable it uses (`conf.OPTIONAL["SIDELOADED_APPS"].items()` means `conf.OPTIONAL["SIDELOADED_APPS"]` is a dict).

Failure is not an issue in itself, but in this case it should happen when the conf is initialized (when `OPTIONAL` is created), rather than when it is used (when the fixture resolves `conf.OPTIONAL[<whatever>]`). This can't be done with a dictionary.

You know I'm a bit biaised toward `dataclasses.dataclas` classes, and I think it would be quite fitting in this case.

```python

@dataclass
class OptionalOptions:
    SIDELOADED_APPS: dict = dict()
    SIDELOADED_APPS_DIR: str = str()
    BACKEND_SCOPE: str = "class"

OPTIONAL = OptionalOptions()
```
... or rather, if you want to explicit to the user where it should be changed:

```python
OPTIONAL = OptionalOptions(
    # use this parameter if you want Speculos to emulate other applications being installed on the device
    # example: {"bitcoin": "Bitcoin", "ethereum": "Ethereum"}
    SIDELOADED_APPS = dict(),
    # Mandatory in case SIDELOADED_APPS is used
    SIDELOADED_APPS_DIR = str(),
    # As the backend instantiation may take some time, Ragger supports multiple backend scopes.
    # You can choose to share the backend instance between {session / module / class / function}
    # When using "session" all your tests will share a single backend instance (faster)
    # When using "function" each test will have its independent backend instance (no collusion)
    BACKEND_SCOPE = "class"
)
```
(and you hide the `class OptionalOptions` declaration in another module).

This way you also get rid of all these "SIDELOADED_APPS" strings buried in the code, which are a pain to maintain, contrary to explicit class properties which will be properly detected/retrieved/corrected by most IDEs.

tests/unit/utils/test_misc.py
```diff
@@ -1,56 +1,37 @@
...
-            filee.unlink()
-        dir_path.rmdir()
-
-    def test_app_path_from_app_name_ok(self):
-        with self.directory() as dir_path:
-            app_name, device = "some_name", "some_device"
-            expected = (dir_path / f"{app_name}_{device}.elf")
-            expected.touch()
-            self.assertEqual(expected, misc.app_path_from_app_name(dir_path, app_name, device))
+class TestMisc():
```
lpascal-ledger, line 7: Unnecessary `()`

tests/unit/utils/test_misc.py
```diff
@@ -1,56 +1,37 @@
...
 
-    def test_app_path_from_app_name_nok_no_directory(self):
-        with self.assertRaises(AssertionError):
-            misc.app_path_from_app_name(Path("/this/does/not/exists/"), "a", "b")
+    def test_find_project_root_dir_ok(self, root_pytest_dir):
+        assert root_pytest_dir.parent == misc.find_project_root_dir(Path(__file__).parent)
```
lpascal-ledger, line 10: This is an indirect test. It works because the `root_pytest_dir.parent` happens to have a `.git` directory, which is not guaranteed (by the setup of this test at least). So this test could suddenly fail whereas the tested function were not changed at all. 

You should rather craft a custom FS with a `.git` directory to check this feature.

tests/unit/utils/test_misc.py
```diff
@@ -1,56 +1,37 @@
+import pytest
```
lpascal-ledger, line 1: It would be great if the unittests were not dependent to pytest, and in particular, to the pytest fixtures we declare. We want the unittest to run with the less dependencies / arguments possible. `root_pytest_dir` is a very simple fixture, but it is a first step of tying together two worlds we do not want to mix.

src/ragger/conftest/base_conftest.py
fbeutin-ledger: It's not as resilient for example when messing with `None` values or if a boolean is inserted
I find the idea of checking the length of the list more clear too

.github/workflows/build_and_tests.yml
fbeutin-ledger: Normally yes but I got burned too many times by the protobuf issues.
I'll remove it but if it comes back I'll just add the env again I won't debug

template/conftest.py
```diff
@@ -1,156 +1,15 @@
-import pytest
-from typing import Optional
-from pathlib import Path
-from ragger.firmware import Firmware
-from ragger.backend import SpeculosBackend, LedgerCommBackend, LedgerWalletBackend
-from ragger.navigator import NanoNavigator, StaxNavigator
-from ragger.utils import app_path_from_app_name
+from ragger.conftest import configuration
```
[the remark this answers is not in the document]
fbeutin-ledger: Intentional

src/ragger/conftest/configuration.py
fbeutin-ledger: Done

src/ragger/utils/misc.py
fbeutin-ledger: Nice catch about the retrocompatible, I'll add it back

.github/workflows/build_and_tests.yml
xchapron-ledger: Deal :)

xchapron-ledger (commented):
Can you update the Changelog too, this is a great feature that need to be in there!

template/conftest.py
fbeutin-ledger: Mmh not a bad idea but I would prefer linking to a documentation than a source file.. How about merging this version and we improve after ?

src/ragger/conftest/base_conftest.py
fbeutin-ledger: Done

src/ragger/conftest/base_conftest.py
fbeutin-ledger: Done

src/ragger/conftest/base_conftest.py
fbeutin-ledger: Done

src/ragger/conftest/configuration.py
fbeutin-ledger: Ok

src/ragger/conftest/configuration.py
fbeutin-ledger: Done with option 2 but I did not hide it in another module, I'm not sure it would be useful: the configuration module provides the class type and it's default value
Option 1 did not work I had issues with `dict()` being refused for default assignation or whatever

tests/unit/utils/test_misc.py
fbeutin-ledger: Done

tests/unit/utils/test_misc.py
fbeutin-ledger: Done

tests/unit/utils/test_misc.py
fbeutin-ledger: Reverted the fixture usage

fbeutin-ledger:
> Can you update the Changelog too, this is a great feature that need to be in there!

Done

template/conftest.py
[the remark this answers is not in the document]
fbeutin-ledger: It's to showcase how to import and use the conf
If the user tries to define a new conf class and modify it's values it won't work for instance
Yes it might pop up in every repo that uses the template but they can very well remove this line in their code

lpascal-ledger (approved):

src/ragger/conftest/configuration.py
lpascal-ledger: Great, thanks!

lpascal-ledger (approved):

lpascal-ledger (approved):

PR #96, opened 2023-01-25
Update conftest and fix use_only_on_backend fixture

fbeutin-ledger:


template/conftest.py
```diff
@@ -136,11 +136,11 @@ def navigator(backend, firmware, golden_run):
         raise ValueError(f"Device '{firmware.device}' is unsupported.")
 
 
-@pytest.fixture(autouse=True)
-def use_only_on_backend(request, backend):
+@pytest.fixture
+def use_only_on_backend(request, backend_name):
```
xchapron-ledger, line 140: Is this one really used?

tests/conftest.py
```diff
@@ -1,71 +1,150 @@
...
-    if backend.lower() == "ledgercomm":
-        return LedgerCommBackend(application.firmware, interface="hid")
-    elif backend.lower() == "ledgerwallet":
-        return LedgerWalletBackend(application.firmware)
-    elif backend.lower() == "speculos":
-        assert application.path.is_file(), f"'{application.path}' elf must exist"
-        return SpeculosBackend(application.path, application.firmware)
-    else:
-        raise ValueError(f"Backend '{backend}' is unknown. Valid backends are: {BACKENDS}")
+@pytest.fixture(scope="session")
```
xchapron-ledger, line 44: Some option doesn't really apply for this tests/conftest.py But I guess you are adding them so that you can latter use the configuration conftest?

xchapron-ledger (dismissed):

template/conftest.py
fbeutin-ledger: In `one` test in ragger, nowhere else, that's why we only see the issue now.

That being said this fixture makes sense so I suggest keeping it

tests/conftest.py
fbeutin-ledger: Yes, in the next PR this conftest is reduced to 10 lines !

After all if a test doesn't want a fixture he doesn't subscribe itself to it and there is no overhead

PR #98, opened 2023-02-07
Update to new speculos detecting elf metadata and allowing nanosp 1.0.4 and stax

xchapron-ledger:


.github/workflows/build_and_tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     - name: Clone
       uses: actions/checkout@v3
     - name: Speculos dependencies
-      run: sudo apt-get update && sudo apt-get install -y qemu-user-static
```
xchapron-ledger, line 23: @lpascal-ledger @fbeutin-ledger Do you known why we are not using Speculos docker image here?

fbeutin-ledger: No idea

xchapron-ledger: Temporally resolved by a fix on Speculos. Should be fixed later too.

fbeutin-ledger (approved):

PR #106, opened 2023-02-27
Misc fix

xchapron-ledger:


src/ragger/backend/speculos.py
```diff
@@ -80,6 +80,20 @@ def __init__(self,
     def url(self) -> str:
         return f"http://{self._host}:{self._port}"
 
+    def _retrieve_client_screen_content(self) -> dict:
+        raw_content = self._client.get_current_screen_content()
```
agrojean-ledger, line 84: Hopefully some OCR improvements should prevent the need for this kind of work around

tests/functional/test_boilerplate.py
```diff
@@ -28,14 +29,21 @@ def test_error_raises_not_returns(backend):
...
+        backend.get_current_screen_content()
+
         for _ in range(right_clicks[firmware.device]):
             backend.right_click()
+            backend.wait_for_screen_change()
 
         with pytest.raises(ConnectionError):
             # clicking on "Quit", Speculos then stops and raises
             backend.both_click()
+            time.sleep(1)
```
agrojean-ledger, line 41: Why the sleep ?

agrojean-ledger (dismissed):

tests/functional/test_boilerplate.py
xchapron-ledger: To wait for speculos to process the both_click action which ends-up in an app quit.
Previously this was done under the hood in the click. It's still the case but I think we shouldn't rely on it.

agrojean-ledger (approved):

PR #108, opened 2023-03-01
Fixes Alpine builds

apaillier-ledger:
Updates the version of the `py-sr25519-bindings` dependency which was causing Rust building issues on Alpine.


xchapron-ledger (dismissed):
why not a `>=`?

apaillier-ledger:
> why not a `>=`?

@xchapron-ledger 
Indeed, I changed it to match 0.2.X instead

fbeutin-ledger (approved):

PR #111, opened 2023-03-15
Update/doc

lpascal-ledger:


doc/glossary.rst
```diff
@@ -77,15 +90,27 @@ Glossary
...
 
    RAPDU
-     "Response APDU", designates the response of an :term:`application`
-     following an :term:`APDU` from the :term:`client` to the application.
+     "Response APDU" (**RAPDU**), designates the response of an
+     :term:`application` following an :term:`APDU` from the :term:`client` to
+     the application.
 
    SDK
      The **SDK** is the open-source code allowing an application to be compiled
```
agrojean-ledger, line 101: The SDK also provides libraries and other helpful things to write apps. Maybe add a short sentence on that @lpascal-ledger ?

doc/tutorial_conftest.rst
```diff
@@ -0,0 +1,528 @@
+Using and understanding the basis of ``Ragger`` testing capabilities
+====================================================================
+
+``Ragger`` is mostly used as a testing framework for Ledger application. To ease
+the integration in such scenario, it includes pre-defined ``pytest`` fixtures and
```
agrojean-ledger, line 5: What are other usage scenarios for ragger ? Not sure we have to say it's "mostly" used as a testing framework... For now, it seems to me it is "only" used for testing, so...

doc/glossary.rst
lpascal-ledger: Yes indeed!

doc/tutorial_conftest.rst
lpascal-ledger: Ragger is at first a client framework, which happens to be well fitted for testing and provides capabilities in this direction. I don't want to corner it in this latter role only.

doc/glossary.rst
lpascal-ledger: Added this sentence:

> On top of an interface to exploit :term:`BOLOS` capabilities, it provides boilerplate functions, graphic abstractions and other useful libraries for developing apps.

xchapron-ledger (approved):

doc/tutorial_conftest.rst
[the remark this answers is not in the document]
adrienlacombe-ledger: indeed, I still see some plugins being submitted with broken snapshots!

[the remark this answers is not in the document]
lpascal-ledger: There [is already a part on this.](https://ledgerhq.github.io/ragger/tutorial_conftest.html#assessing-that-the-test-is-right).
It's not a warning, but still a dedicated section of the tutorial.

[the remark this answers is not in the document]
lpascal-ledger: I don't think so, this is out of `Ragger` scope, but we could emphasize the usage of `pytest` and link to its own documentation.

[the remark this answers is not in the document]
lpascal-ledger: I'll add follow-up improvment in the #113  PR

PR #119, opened 2023-05-04
Misc change to improve ledgerwallet tests

xchapron-ledger:


src/ragger/backend/interface.py
```diff
@@ -77,6 +77,16 @@ def __exit__(self, exc_type: Optional[Type[BaseException]], exc_val: Optional[Ba
                  exc_tb: Optional[TracebackType]):
         raise NotImplementedError
 
+    def handle_usb_reset(self) -> None:
+        """
+        Inform the backend that is should handle an USB reset.
```
fbeutin-ledger, line 82:
```suggestion
        Inform the backend that it should handle an USB reset.
```

src/ragger/backend/ledgercomm.py
```diff
@@ -69,6 +69,9 @@ def __exit__(self, *args, **kwargs):
         assert self._client is not None
         self._client.close()
 
+    def handle_usb_reset(self) -> None:
+        pass
```
fbeutin-ledger, line 73: Maybe add a comment to use ledgerwallet here

src/ragger/conftest/base_conftest.py
```diff
@@ -170,6 +171,17 @@ def create_backend(root_pytest_dir: Path, backend_name: str, firmware: Firmware,
 def backend(root_pytest_dir, backend_name, firmware, display, log_apdu_file, cli_user_seed):
     with create_backend(root_pytest_dir, backend_name, firmware, display, log_apdu_file,
                         cli_user_seed) as b:
+        if backend_name.lower() == "ledgerwallet" and conf.OPTIONAL.APP_NAME:
```
fbeutin-ledger, line 174: I'm starting to wonder why we keep `ledgercomm` as we never use it and don't want to maintain both... I guess we'll let ledgercomm live until we pull the plug one day

src/ragger/conftest/configuration.py
```diff
@@ -3,13 +3,17 @@
...
 class OptionalOptions:
+    APP_NAME: str
     SIDELOADED_APPS: dict
     SIDELOADED_APPS_DIR: str
     BACKEND_SCOPE: str
     CUSTOM_SEED: str
 
 
 OPTIONAL = OptionalOptions(
+    # Use this parameter if you want LedgerWallet to make sure the right app is started upon backend instantiation.
```
fbeutin-ledger, line 14: `+` Only used on ledgerwallet backend

src/ragger/conftest/base_conftest.py
xchapron-ledger: Yep agreed, but dropping it is not for this PR :)

src/ragger/conftest/base_conftest.py
```diff
@@ -170,6 +171,17 @@ def create_backend(root_pytest_dir: Path, backend_name: str, firmware: Firmware,
...
+            # Make sure the app is restarted as this is what is requested by the fixture scope
+            app_name, version = get_current_app_name_and_version(b)
+            requested_app = conf.OPTIONAL.APP_NAME
+            if app_name == requested_app:
+                exit_current_app(b)
+                b.handle_usb_reset()
```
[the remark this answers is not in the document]
xchapron-ledger: As discussed, this should be kept, App that don't need it should use a better scope than function scope to avoid having to many app consent to give.
And signing the app is still a possibility!

src/ragger/backend/ledgercomm.py
xchapron-ledger: Eventually I implemented it on ledgercomm too. So for now both are still supporting the same features.

fbeutin-ledger (approved):

PR #123, opened 2023-05-24
Change deployment trigger from commit message to tag

lpascal-ledger:
Remarks:

- tags and version in the CHANGELOG must be equals. This is currently not the case, as tags are like `v1.7.3` while CHANGELOG has `1.7.3`
- previous behavior (the packaging is done - but not pushed - on push and PR, just to check if it works) is reproduced, with additional steps fetching version / branch info and displaying them

### DIFF
--- a/.github/workflows/build_and_tests.yml
+++ b/.github/workflows/build_and_tests.yml
@@ -1,14 +1,11 @@
-name: Build, test and deploy Ragger
+name: Build and test Ragger
 
 on:
   workflow_dispatch:
   push:
     branches:
       - master
-      - main
       - develop
-    tags:
-      - '*'
   pull_request:
 
 jobs:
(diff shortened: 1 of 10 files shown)


.github/workflows/build_and_tests.yml
```diff
@@ -63,39 +65,77 @@ jobs:
...
         python -m twine check dist/*
 
+    - name: Get tag version
+      if: success() && github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/')
+      id: tag
+      run: |
+        echo "TAG_VERSION=${GITHUB_REF#refs/tags/}" >> "$GITHUB_ENV"
+        echo "Tag version: '${{ env.TAG_VERSION }}'"
+
+    - name: Check version against CHANGELOG
```
xchapron-ledger, line 81: Maybe we should also check against `__version__` from `src/ragger/__init__.py`?

lpascal-ledger: No need if the version is automatically updated from the tag version

xchapron-ledger:
I'm wondering what will happen if a version is first tagged on develop, and then develop is merged on master. I'm afraid that won't trigger a deployment on `pypi.org`?

xchapron-ledger:
What's the version format on real tags ? I see `1.7.4.dev15+g1f1edf6` on https://github.com/LedgerHQ/ragger/actions/runs/5070718156/jobs/9106135777#step:6:18  which is great for development, but not to be published.

xchapron-ledger:
Also I'm wondering if the deploy job should really be triggered on pull request and branch merge, this doesn't bring much no?

xchapron-ledger:
>     * tag on `develop` then merged on `master` -> it won't deploy, as the trigger event will not be a tag. But if we simply merge from `develop` to `master` and the only difference (as `master` is just updated from `develop` and nowhere else) is the merge commit, we could change the tag from `develop` to `master`, which will trigger a `pypi.org` deployment. What do you think?

I'm not found of deleting tags to push them again.
What do you thing of using two convention for tags? One prefixed with dev (for deployment on t`est.pypi`), and the other without? This would remove the branch restriction? Also I'm a bit worried of this condition `startsWith(github.ref, 'refs/tags/') && env.BRANCH_NAME == 'origin/develop'` as the commit might be on multiple branch?

>     * `setuptools_scm` calculates version following [this scheme](https://github.com/pypa/setuptools_scm#default-versioning-scheme). In this case, as the last tag (`v1.7.3`) was a few commits ago, the chosen version scheme is `distance and clean: {next_version}.dev{distance}+{scm letter}{revision hash}`, hence the `1.7.4.dev15+g1f1edf6`. If `no distance and clean` (i.e tag is on latest commit), the version will be the tag.

I guess all these version schemes are allowed by pypi?

lpascal-ledger:
> I guess all these version schemes are allowed by pypi?
I think so, as it's regular semantic versionning, and I've already seen `0.1.1.dev19` or other packages on `pypi.org`.
But as in our case the package is deployed on tag, the git state will be clean with no distance, so `setuptools_scm` will exactly match the tag. We could change our tag scheme and break `pypi` deployment, but it would be our doing, not `setuptools_scm`'s.

lpascal-ledger:
> I'm not found of deleting tags to push them again.

Me neither in general.

> What do you thing of using two convention for tags? One prefixed with dev (for deployment on test.pypi), and the other without? This would remove the branch restriction? Also I'm a bit worried of this condition startsWith(github.ref, 'refs/tags/') && env.BRANCH_NAME == 'origin/develop' as the commit might be on multiple branch?

It's a possibility, that would remove the check on the branch. That would also require extra customization of `setuptools_scm` and the workflow to extract the correct. I'll try this.

lpascal-ledger:
@xchapron-ledger implemented the system with tag telling where to deploy.

I went with two basic options:
- normal tag (`1.3.6`) -> mypi.org
- tag precede with `test-`" (`test-1.3.6`) -> test.pypi.org

Version can start with `v`: `v1.3.6` or `test-v1.3.6` are valid.

No more branch check in play.

Example [here](https://github.com/LedgerHQ/ragger/actions/runs/5080784164/jobs/9128224504) with tag `test-9.9.9`. Of course it fails when checking with the CHANGELOG, but the step before `Display current status` shows all the variables to be used for deployment.

lpascal-ledger:
Okay, enforcing the `v` as it has been the habit until the `v0.4.0`.

lpascal-ledger:
The only thing that saddens me is that the regex to extract the prefix `test-` or the version is stored at two locations:
- the `tools/info_extractor.py` file (used by the workflow)
- the `pyproject.toml` file (used by `setuptools_scm`)

I have not found how to have a centralized storage.

xchapron-ledger:
> The only thing that saddens me is that the regex to extract the prefix `test-` or the version is stored at two locations:
> 
>     * the `tools/info_extractor.py` file (used by the workflow)
> 
>     * the `pyproject.toml` file (used by `setuptools_scm`)
> 
> 
> I have not found how to have a centralized storage.

the regexp is not going to change anytime soon, I think having it at 2 places is not such a big issue.

xchapron-ledger (approved):

PR #132, opened 2023-06-07
Handle Speculos not replying outdated events while it performs OCR

fbeutin-ledger:


src/ragger/backend/speculos.py
```diff
@@ -161,6 +167,7 @@ def both_click(self) -> None:
         self._client.press_and_release("both")
 
     def finger_touch(self, x: int = 0, y: int = 0, delay: float = 0.5) -> None:
+        self.logger.info(f"Sending finger_touch x={x}, y={y}, delay={delay}")
```
xchapron-ledger, line 170: Maybe we should add an equivalent logs for button press?

src/ragger/backend/speculos.py
```diff
@@ -82,17 +82,23 @@ def url(self) -> str:
         return f"http://{self._host}:{self._port}"
 
     def _retrieve_client_screen_content(self) -> dict:
-        raw_content = self._client.get_current_screen_content()
+        raw_events = self._client.get_current_screen_content().get("events", [])
+        while raw_events == []:
```
xchapron-ledger, line 86: What if the screen doesn't have text content? Like only glyphs ?

src/ragger/backend/speculos.py
```diff
@@ -82,17 +82,23 @@ def url(self) -> str:
...
+            print(f"raw_events = '{raw_events}'")
```
lpascal-ledger, line 87: To be cleaned?

src/ragger/backend/speculos.py
```diff
@@ -82,17 +82,23 @@ def url(self) -> str:
...
         # They probably comes from long press progress bar on Stax
         # and if not removed they falsely make wait_for_screen_change
         # consider screen as changed when screen text didn't and that's
         # what we want here.
         events = []
-        for event in raw_content.get("events", []):
+        for event in raw_events:
             if event.get("text", "").strip():
                 events.append(event)
+        self.logger.info(f"Received '{events}'")
```
lpascal-ledger, line 101: Same as on Speculos: `.debug` rather? And f-strings.

src/ragger/backend/speculos.py
lpascal-ledger, line 170: Same as before: `.debug` rather? And f-strings.

fbeutin-ledger:
> Is this PR still relevant?

Nope

PR #138, opened 2023-06-28
Removing firmware version mechanism

lpascal-ledger:
Until now, Ragger was keeping track of firmware version for two usages:

- being able to properly start Speculos (`--model nanos --sdk 2.1.0`)
- keeping track of graphical object positions in Stax

The first need is now resolved for quite some time as the version is embedded into the application ELF files, hence Speculos figures it out alone.
The second one is not aligned with Ledger policy which for now only supports the **latest** firmware version. Maintaining the current mechanism for just this purpose is probably not worth it (and would be marginally useful, as it is expected that graphical movement in Stax should be quite rare).

So this PR removes the version mechanism, then simplify the `Firmware` class, which is now a simple enumerate with some fancy properties on top.

Not sure on the version bump. It should be pretty invisible for those who rely on the `ragger.conftest` module, but anyone crafting custom `Firmware` will be impacted. As it could be a breaking change for them, I defaulted on a major version but :shrug:


xchapron-ledger (dismissed):
Not sure it deserve a Major bump, but after all, why not.

lpascal-ledger:
Moved to a minor bump.
The change is hardly breaking, and would only impact rare users (only myself probably).

PR #142, opened 2023-07-18
Improve test_name fixture by only removing the 'firmware' parameter

fbeutin-ledger:


xchapron-ledger (commented):
Looks interesting but this will break some snapshot location on some app :/
Might still be worth it.

fbeutin-ledger:
> Looks interesting but this will break some snapshot location on some app :/ Might still be worth it.
I know, but it's needed for this:
https://github.com/LedgerHQ/app-exchange/blob/fbe/exchange_test_runner/test/python/test_stellar.py

Currently our tests using parametrization cannot use the navigation which is unfortunate
OR they can modify the `test_name` locally 🤔

xchapron-ledger:
> > Looks interesting but this will break some snapshot location on some app :/ Might still be worth it.
> > I know, but it's needed for this:
> > https://github.com/LedgerHQ/app-exchange/blob/fbe/exchange_test_runner/test/python/test_stellar.py
> 
> Currently our tests using parametrization cannot use the navigation which is unfortunate OR they can modify the `test_name` locally thinking

Yep, that's what I'm using here: https://github.com/LedgerHQ/app-eos/blob/ea123cd6d9ac7ec8f24ba0a34e3073badaf735e5/tests/functional/test_sign_cmd.py#L50-L53
But I can change it obviously.

fbeutin-ledger:
Dropped, to handle in tests

PR #143, opened 2023-07-19
Add all_nano option to the device parameter in the base conftest

fbeutin-ledger:


src/ragger/conftest/base_conftest.py
```diff
@@ -91,21 +91,14 @@ def pytest_generate_tests(metafunc):
...
 
-        if device == "all":
-            if backend_name != "speculos":
-                raise ValueError("Invalid device parameter on this backend")
-
-            # Add all supported firmwares
-            for fw in FIRMWARES:
+        # Enable firmware for requested devices
+        for fw in FIRMWARES:
+            if (device == fw.name) or (device == "all") or (device == "all_nano" and fw.name.startswith("nano")):
```
xchapron-ledger, line 96: I think you could use https://github.com/LedgerHQ/ragger/blob/master/src/ragger/firmware/structs.py#L41 ?

xchapron-ledger (approved):

PR #155, opened 2023-11-08
interface: add RAISE_CUSTOM policy

sgliner-ledger:


src/ragger/backend/interface.py
```diff
@@ -30,11 +30,15 @@ class RaisePolicy(Enum):
...
+    RAISE_CUSTOM = auto()
 
 
 class BackendInterface(ABC):
 
-    def __init__(self, firmware: Firmware, log_apdu_file: Optional[Path] = None):
+    def __init__(self,
+                 firmware: Firmware,
+                 log_apdu_file: Optional[Path] = None,
+                 whitelisted_status: Optional[list] = None):
```
lpascal-ledger, line 41: If the default value is a list, might as well defining it here. It will simplify the following code.
```suggestion
                 whitelisted_status: list = []):
```

src/ragger/backend/interface.py
```diff
@@ -50,6 +54,11 @@ def __init__(self, firmware: Firmware, log_apdu_file: Optional[Path] = None):
         self.logger = get_default_logger()
         self.apdu_logger = get_apdu_logger()
 
+        if whitelisted_status:
+            self.whitelisted_status = whitelisted_status
+        else:
+            self.whitelisted_status = []
+
```
lpascal-ledger, line 61: If previous suggestion accepted:
```suggestion
        self.whitelisted_status = whitelisted_status
 ```

**Review by lpascal-ledger — APPROVED** [review]
This exception mechanism lacks unit tests. Might be a good occasion?

File: src/ragger/backend/interface.py (hunk 1 above)
Comment (sgliner-ledger) [reply src/ragger/backend/interface.py:41]: I did that at first, but I had a codesmell detected :(

Comment (lpascal-ledger) [reply src/ragger/backend/interface.py:41]: Probably because of dangerous side-effect, but we don't trigger that pitfall.
Don't know how to bypass though :thinking:

Comment (lpascal-ledger) [reply src/ragger/backend/interface.py:41]: Maybe by using tuples instead of list? It stays immutable, that's consistent with our usage.

Comment (lpascal-ledger) [reply src/ragger/backend/interface.py:41]:
```suggestion
                 whitelisted_status: Iterable = ()):
```

sgliner-ledger: It was somehting about setting the default value of a variable in a function being bad practice for maintenability :thinking:

src/ragger/backend/interface.py
sgliner-ledger: :heavy_check_mark:

lpascal-ledger (approved):

PR #157, opened 2023-12-21
Batch SpeculosBackend clients

lpascal-ledger:
## What it does

This feature allows to declare several Speculos clients at once.

It is useful for testing an application which works with several instances of itself.

Usage example:
```python
@pytest.fixture
def clients():
    from ragger.backend import SpeculosBackend
    from ragger.firmware import Firmware
    s1, s2 = SpeculosBackend.batch('build/nanos/bin/app.elf',
                                   Firmware.NANOS,
                                   2,
                                   args=["--display", "qt"])
    with s1, s2:
        yield ClientWrapper(s1), ClientWrapper(s2)

def test_hello(clients):
    client1, client2 = clients
    name = client1.get_name()
    client2.cheer(name)
```

## Dependencies

- Needs [Speculos changes](https://github.com/LedgerHQ/speculos/pull/440/files)
- Needs [Ledgered changes](https://github.com/LedgerHQ/ledgered/pull/12/files)

## Other thoughts

At first I was thinking of adding a related fixture in `conftest/base_conftest.py`. However it would require a **lot** of  modification all around the place to get the other relevant fixtures / values (backend, firmware, options, ...), leading to either extensive refactorization and/or having lots of code duplication.
On top of that, the fixture would be only valid for `Speculos`, not the other backends (maybe at some point if we are confident with USB ports handling, but that's a big if, in any case not for now).
And finally it would probably not be used by lots of folks anyway.

So I didn't.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,6 +56,7 @@ doc=
         docutils==0.16  # higher versions trigger build bugs with the RTD theme
 speculos=
         speculos>=0.2.8
+        mnemonic
 ledgercomm=
         ledgercomm
         ledgercomm[hid]
--- a/src/ragger/backend/speculos.py
+++ b/src/ragger/backend/speculos.py
@@ -272,6 +272,17 @@ class SpeculosBackend(BackendInterface):
                 return
 
     @classmethod
+    def clean_args(cls: Type[T], speculos_args: List) -> None:
+        logger = get_default_logger()
+        for argument in [cls._ARGS_APDU_PORT_KEY, cls._ARGS_API_PORT_KEY]:
+            if argument in speculos_args:
+                logger.warning("'%s' argument is ignored on batch mode", argument)
+                index = speculos_args.index(argument)
+                # popipng argument and its value
+                speculos_args.pop(index)
+                speculos_args.pop(index)
+
+    @classmethod
     def batch(cls: Type[T],
               application: Path,
               firmware: Firmware,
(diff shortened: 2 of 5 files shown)


src/ragger/backend/speculos.py
```diff
@@ -250,3 +269,42 @@
...
+            if different_seeds:
+                additional_args.extend(["--seed", generate_mnemonic()])
+            if different_rng:
+                additional_args.extend(["--deterministic-rng", f"{apdu_port}{api_port}"])
+            if different_private:
+                additional_args.extend(["--user-private-key", urandom(32).hex()])
+            if different_attestation:
+                additional_args.extend(["--attestation-key", urandom(32).hex()])
+            if "args" in tmp_kwargs:
+                tmp_kwargs["args"].extend(additional_args)
```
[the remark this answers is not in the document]
lpascal-ledger: False positive? `tmp_kwargs = deepcopy(kwargs)`, so `kwargs` shall not be impacted by `tmp_kwargs` modification.

src/ragger/backend/speculos.py
```diff
@@ -250,3 +269,42 @@
...
+            else:
+                tmp_kwargs["args"] = additional_args
```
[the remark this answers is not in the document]
lpascal-ledger: Same as above.

setup.cfg
```diff
@@ -56,6 +56,7 @@ doc=
         docutils==0.16  # higher versions trigger build bugs with the RTD theme
 speculos=
         speculos>=0.2.8
```
xchapron-ledger, line 58: Not really related to this PR, but would be great to bump this speculos version and to remove all references to pytesseract

xchapron-ledger: Same below in `all_backends` part

xchapron-ledger (approved):

setup.cfg
lpascal-ledger: Indeed, will clean this in another PR soon.

PR #167, opened 2024-02-15
backend: Add optional tick_timeout argument to exchange functions

sgliner-ledger:


xchapron-ledger (commented):
Maybe the default value can be changed?
Also can you bump the speculos required version?

xchapron-ledger:
Can you bump the required speculos version?

pyproject.toml
```diff
@@ -61,7 +61,7 @@ doc = [
     "docutils==0.16",
 ]
 speculos = [
-    "speculos>=0.5.0",
```
xchapron-ledger, line 64: Last speculos version is something like 0.7.1, why not using it?

sgliner-ledger: I can change it, but I was thinking the upgrade is not mandatory as it doesnt break compatibility

xchapron-ledger (approved):

PR #172, opened 2024-03-07
Using ledgered as entrypoint to infer Rust & C app binary location

lpascal-ledger:
@yogh333 following #171 I tried this, it seems to work with the boilerplate `C` and `Rust` apps.
Does that seems ok?

(Tests are not converted yet, I expect them to fail)


yogh333 (approved):
LGTM 🥳

src/ragger/utils/misc.py
```diff
@@ -47,10 +48,15 @@ def find_library_application(base_dir: Path, name: str, device: str) -> Path:
...
 
 
-def find_main_application(base_dir: Path, build_dir: str, bin_name: str, device: str) -> Path:
+def find_main_application(base_dir: Path, device: str, sdk: str) -> Path:
     if not base_dir.is_dir():
         raise AssertionError(f"{base_dir} is not a directory")
-    app = Path(base_dir / build_dir / device / bin_name).resolve()
+    app = base_dir.resolve()
+    if sdk == "rust":
+        app_name = toml.load(base_dir / "Cargo.toml")["package"]["name"]
```
agrojean-ledger, line 56: Wouldn't it be better if the `ledgered` tool inferred this as well ? It seems like a better place than doing in it in Ragger.

fbeutin-ledger:
Hard to understand the output of the two PRs
Can we merge this one in https://github.com/LedgerHQ/ragger/pull/171/files ?

fbeutin-ledger:
Please do not merge the combination until I've tested with plugins 👍

yogh333:
> Hard to understand the output of the two PRs Can we merge this one in https://github.com/LedgerHQ/ragger/pull/171/files ?

No, this PR is a better proposal to manage what #171 is supposed to.

src/ragger/utils/misc.py
yogh333: @agrojean-ledger Actually, it is already the case: `sdk` string is set from `manifest.app.sdk` passed as argument to `find_main_application` function.

fbeutin-ledger:
Capture
This PR is not independent, it should be merged into the target branch so we can see the real diff

xchapron-ledger:
Or just change this PR to target develop.

fbeutin-ledger:
Ah j'ai lu l'inverse j'ai cru qu'elle partait de jca/manage_rust_buildtree

src/ragger/utils/misc.py
lpascal-ledger: @agrojean-ledger Indeed there is quite some functions in this `utils` module which could be internalized into `ledgered`, including this one. Let's list that into [this issue](https://github.com/LedgerHQ/ledgered/issues/11) until I got time.

lpascal-ledger:
I've updated the branch so that it is backed against `master`. It should be up-to-date and testable now @fbeutin-ledger 
I'll also update the tests accordingly.

tests/unit/utils/test_misc.py
```diff
@@ -50,7 +53,8 @@
             tmp_dir.mkdir(parents=True, exist_ok=True)
             expected = tmp_dir / "app.elf"
             with self.assertRaises(AssertionError) as error:
-                misc.find_main_application(dir_path, device)
+                misc.find_main_application(dir_path, conf.OPTIONAL.BUILD_DIR,
+                                           conf.OPTIONAL.BIN_NAME, device)
```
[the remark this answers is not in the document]
lpascal-ledger: We've cornered an error with the current system when using the `LOAD_MAIN_APP_AS_LIBRARY` option.
This option is used when the `MAIN_APP` (the app from the current repository) is to be loaded as a library. In this case, the `app_path` refers to another app which will be the app started by Speculos. 

This is the case for eth plugins:
- the tests run from the plugin repository, the plugin is the main app
- `app_path` will point into the ETH app binary

This highlights an ambiguous usage of the `APP_DIR` variable. 
- On most cases it is the repository of the current application, and its value is either `.` (default value) or `./app` (Zondax app). This becomes redundant with manifests as it is exactly the `app.build_directory` value.
- On plugins cases it becomes the ETH app binary root directory. This is the buggy case.

We agreed on the following:
- using the qualifier `MAIN_APP` to designate the application which should be run first, instead of designating the application from which repo the tests runs;
- application from which repo the tests runs is designated as `LOCAL_APP`;
- A new optional value `MAIN_APP_DIR` is created, which if not None means that:
  - the `LOCAL_APP` is not the `MAIN_APP` and thus shall be loaded as a library (`LOAD_MAIN_APP_AS_LIBRARY` becomes obsolete)
  - the `APP_DIR` of this foreign `MAIN_APP` is this `MAIN_APP_DIR` directory.

This to our mind will make this part clearer.

An issue remains: without the `MAIN_APP` manifest, it is not possible to know if the `MAIN_APP` is a Rust or C application, and from now inferring the proper binary path is not possible.
As there is no plan to have applications running as libraries of a Rust application, we thought this could be ignored for now, and that we could assume every `MAIN_APP` which are not `LOCAL_APP` are C applications.

lpascal-ledger:
With @fbeutin-ledger we've cornered an error with the current system when using the `LOAD_MAIN_APP_AS_LIBRARY` option.
This option is used when the `MAIN_APP` (the app from the current repository) is to be loaded as a library. In this case, the `app_path` refers to another app which will be the app started by Speculos. 

This is the case for eth plugins:
- the tests run from the plugin repository, the plugin is the main app
- `app_path` will point into the ETH app binary

This highlights an ambiguous usage of the `APP_DIR` variable. 
- On most cases it is the repository of the current application, and its value is either `.` (default value) or `./app` (Zondax app). This becomes redundant with manifests as it is exactly the `app.build_directory` value.
- On plugins cases it becomes the ETH app binary root directory. This is the buggy case.

We agreed on the following:
- using the qualifier `MAIN_APP` to designate the application which **should be run first** (the app which is **not** a library), instead of designating the application from which repo the tests runs;
- application from which repo the tests runs is designated as `LOCAL_APP`;
- A new optional value `MAIN_APP_DIR` is created, which if not None means that:
  - the `LOCAL_APP` is not the `MAIN_APP` and thus shall be loaded as a library (`LOAD_MAIN_APP_AS_LIBRARY` becomes obsolete)
  - the `APP_DIR` of this foreign `MAIN_APP` is this `MAIN_APP_DIR` directory.

This to our mind will make this part clearer.

An issue remains: when `LOCAL_APP` is to be loaded as a library, we don't have access to the `MAIN_APP` manifest; so it is not possible to know if the `MAIN_APP` is a Rust or C application, and so inferring the proper binary path is not possible.
As there is no plan to have applications running as libraries of a Rust application, we thought this could be ignored for now, and that we could assume every `MAIN_APP` which are not `LOCAL_APP` are C applications.

yogh333:
> An issue remains: when `LOCAL_APP` is to be loaded as a library, we don't have access to the `MAIN_APP` manifest; so it is not possible to know if the `MAIN_APP` is a Rust or C application, and so inferring the proper binary path is not possible.
> As there is no plan to have applications running as libraries of a Rust application, we thought this could be ignored for now, and that we could assume every `MAIN_APP` which are not `LOCAL_APP` are C applications.

When testing a plugin for Ethereum app, how is the Ethereum app provided ? As a pre-built binary ? If it is the case, providing the `elf` binaries in the proper folder should be enough (e.g. `nanos/app-ethereum.elf`, `nanosplus/app-ethereum.elf`, etc...) no matter if these binaries are generated from the C-SDK or Rust-SDK toolchain. Why do we need to get the full build tree for the main app ? Except if the main application build is part of the testing process of a plugin, but it does not seem to be the case. 🤔

lpascal-ledger:
It is not so much that we _need_ to, but that it is the current behavior, and so breaking it should be done cautiously.
Changing that would require changes into the workflows, the VScode extension and the apps using this feature. I'm not sure we want to integrate this change in this PR right now.

lpascal-ledger:
@yogh333 @fbeutin-ledger PR ready, waiting for your tests!

lpascal-ledger:
@yogh333 weird, this bug is related to this merged PR #167, it should not trigger if Ragger is properly installed.
What's your repo so that I could check?

yogh333:
Ok after updating my Docker container, it is better.
Only one issue with nanosplus device name:
```
pytest tests/ --tb=short -v --device nanosplus
[2024-03-11 16:10:43,250][DEBUG] ragger.logger - Running Ragger with the following conf:
[2024-03-11 16:10:43,250][DEBUG] ragger.logger - 	APP_NAME = ''
[2024-03-11 16:10:43,250][DEBUG] ragger.logger - 	MAIN_APP_DIR = 'None'
[2024-03-11 16:10:43,250][DEBUG] ragger.logger - 	SIDELOADED_APPS = '{}'
[2024-03-11 16:10:43,250][DEBUG] ragger.logger - 	SIDELOADED_APPS_DIR = ''
[2024-03-11 16:10:43,250][DEBUG] ragger.logger - 	BACKEND_SCOPE = 'class'
[2024-03-11 16:10:43,250][DEBUG] ragger.logger - 	CUSTOM_SEED = ''
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: argument --device: invalid choice: 'nanosplus' (choose from 'nanos', 'nanox', 'nanosp', 'stax', 'all', 'all_nano')
```

lpascal-ledger:
We will **not** add a duplicate NanoS+ target as a valid Ragger device. That would be pure confusion from a user perspective. Either the Rust builds aligns with the `nanosp` naming, or Ragger internally converts it to the relevant value.

yogh333:
> We will **not** add a duplicate NanoS+ target as a valid Ragger device. That would be pure confusion from a user perspective. Either the Rust build aligns with the `nanosp` naming, or Ragger internally converts it to the relevant value.

ok let me check, it should be handled by Rust

yogh333:
When building apps based on Rust SDK, the output build tree is stored in `target/_device_/release` where `_device_` is `nanos`, `nanosplus` or `nanox` (`stax` and `europa` soon). For NanoS+, the build command is `cargo ledger build nanosplus` and the binaries will be produced in `target/nanosplus/release`. In Rust SDK, it is not a big deal to change the NanoS+ target from `nanosplus` to `nanosp` (`cargo ledger build nanosp`) and in that case the binaries will be stored in `target/nanosp/release`. But a modification is still required in Ragger as `nanosp` string is systematically replaced by `nanos2` and generating the binaries into `target/nanos2/release` when invoking `cargo ledger build nanosp` is not consistent.

lpascal-ledger:
So the Rust directory name is (will be?) `nanosplus`? Confirmed?

yogh333 (approved):
LGTM

PR #179, opened 2024-04-04
navigator.py: Add use case navigation function

xchapron-ledger:


lpascal-ledger:
Thanks. I was overthinking the issue and was skeptical as for a simple implementation.

This solution highlights a repetition repetition that could be factorized - not for engineering faffing, but to anticipate on futur additions which will surely be plenty every time a new "use case scenario" (or a new device) is integrated.

As always when the code is redundant except for the variables, we should decouple the data from the processing (that's actually why the "navigation" is separated from the "firmware").

To be more explicit, I would go for a pure declarative definition of these scenarios (complex dicts are evil and dataclass are often better but let's think of it as pseudo-code):
```
Enum Scenario
Enum ExpectingType
scenario_scripts = { Scenario.REVIEW: {
    Firmware.NANO: {  
        "navigation": NavInsID.RIGHT_CLICK,
        "validation": [NavInsID.BOTH_CLICK]
        "expecting": {ExpectingType.APPROVE: "Approve", ExpectingType.REJECT: "Reject" }
    },
    Firmware.STAX: { 
        "navigation": NavInsID.USE_CASE_REVIEW_TAP,
        "validation": [NavInsID.USE_CASE_ADDRESS_CONFIRMATION_CONFIRM, NavInsID.USE_CASE_STATUS_DISMISS]
        "expecting": {ExpectingType.APPROVE: "Approve", ExpectingType.REJECT: "Reject" }
    },
},
    Scenario.ADDRESS_REVIEW : {
        [...]
    },
}
```
... then no need for several methods, only one:

```
def run_scenario(self,
                               scenario: Scenario,
                               path: Optional[Path] = None,
                               test_case_name: Optional[Path] = None,
                               expecting_type: ExpectingType = ExpectingType.APPROVE,
                               expecting_text: Optional[str] = None):
    if expecting_type != ExpectingType.CUSTOM and expecting_text is not None:
        raise ValueError("expecting text is ignored")
    if expecting_type == ExpectingType.CUSTOM and expecting_text is None:
        raise ValueError("")
    script = scenario_scripts[scenario]
    if script.get(self._firmware.device) is None:
        raise NotImplementedError("Device not supported")
    navigate_instruction = script[self._firmware.device]["navigation"]
    validation_instructions = script[self._firmware.device]["validation"]
    if excepting_type == ExpectingType.CUSTOM:
        expecting_screen_text = expecting_text
    else:
        excepting_screen_text = script[self._firmware.device][expecting_type]
    self.navigate_until_text_and_compare(navigate_instruction,
                                             validation_instructions,
                                             approve_screen_text,
                                             path,
                                             test_case_name)
```
It should save some lines, clearly separates the scenario description from the code, allowing to add lots of other scenario or widening existing one very easily. Also it would be a lot easier to test.

Calls would look like:
```
- navigator.address_review_approve(ROOT_SCREENSHOT_PATH, test_name)
+ navigator.run_scenario(Scenario.ADDRESS_REVIEW, ROOT_SCREENSHOT_PATH, test_name)

- navigator.address_review_reject(ROOT_SCREENSHOT_PATH, test_name)
+ navigator.run_scenario(Scenario.ADDRESS_REVIEW, ROOT_SCREENSHOT_PATH, test_name, ExpectingType.REJECT)

- navigator.review_approve(ROOT_SCREENSHOT_PATH, test_name)
+ navigator.run_scenario(Scenario.REVIEW, ROOT_SCREENSHOT_PATH, test_name)

- navigator.review_reject(ROOT_SCREENSHOT_PATH, test_name)
+ navigator.run_scenario(Scenario.REVIEW, ROOT_SCREENSHOT_PATH, test_name, ExpectingType.REJECT)
```

xchapron-ledger:
@lpascal-ledger We could go this path, but this assume that we will only use navigate_until_text_and_compare for the navigation. I'm not sure that at some point we will not want to use something else?

fbeutin-ledger:
TBH I'm not sure I see the benefit of having the scenarios ? While it would complexify the syntax a lot

lpascal-ledger:
> @lpascal-ledger We could go this path, but this assume that we will only use navigate_until_text_and_compare for the navigation. I'm not sure that at some point we will not want to use something else?

Good point. That highlights that:
1. decoupling data from process is even more interesting, as "scenarios" (that's the name I came up with first but there's probably better / clearer) can be processed whatever the finishing method: `navigate_until_text_and_compare` or `navigate_and_compare` or `navigate_until_text`, there's only some data to inject or not in the finishing method but they all basically do the very same thing.
2. the methods in this PR are wrappers around `navigate_until_text_and_compare`. We should reflect this relation in some ways. 
  - We could have a new method for each finishing method (like `run_scenario_until_text_and_compare`, `run_scenario_until_text`, ...) but maybe we are entering too deep in the way_too_long_name_describing_how_the_method_behaves territory.
  -  Allowing the user to give this finishing method as a callback parameter feels a bit clumsy (`navigator.run_scenario(Scenario.REVIEW, ROOT_SCREENSHOT_PATH, test_name, navigator.navigate_until_text_and_compare)`)
  - In the same spirit but maybe a bit better, there could be :anotherone: Enum directing to the correct method (`navigator.run_scenario(Scenario.REVIEW, ROOT_SCREENSHOT_PATH, test_name, Navigator.UNTIL_TEXT_AND_COMPARE)`). That feels heavy though.
  - ??

> TBH I'm not sure I see the benefit of having the scenarios ? While it would complexify the syntax a lot

That depends on what's your complexity measurement. I can think of the following (there may be more):

- It is **more** complex because constants are replaced with variables stored somewhere else (that feels very minor to me. I assume every developer can figure out how to replace a variable with its value depending on the parameter)
- It is **more** complex because there is two more checks than in the previous methods (again, we can manage two `if`s)
- It is **less** complex because there is less code lines
- It is **less** complex because there is only one method to test, instead of `N` methods to test (each one the exact same way), and because it is easier to test as data is immediately mockable.
- It is **less** complex because any new behavior from now on will take place in a single method, instead of being reported in `N` methods with the risk of forgetting or misreporting a change (and the pain to read `N` time the same change in PRs, with the risk of skipping a tiny error in the process just because we are humans)
- It is **less** complex because any new scenario, or new device in an existing scenario is only an entry inside a single data structure, instead of adding a new method (and tests) copy-pasting the others, or reporting the new device in `N` methods with the risk of forgetting or misreporting a change.

My personal experience is that the ease to test code is the best way to measure code complexity. I confess I assess code with this bias.

src/ragger/conftest/base_conftest.py
```diff
@@ -224,6 +226,9 @@ def use_only_on_backend(request, backend_name):
         if current_backend != backend_name:
             pytest.skip(f'skipped on this backend: "{current_backend}"')
 
+@pytest.fixture(autouse="function")
+def default_screenshot_path(request):
+    return request.path.parent.resolve()
```
xchapron-ledger, line 231: Not sure how this one behaves depending on where pytest is launched?
Maybe something based on a ROOT_SCREENSHOT_PATH to be defined in conftest.py or something could be better?

src/ragger/navigator/navigation_scenario.py
```diff
@@ -0,0 +1,82 @@
...
+from typing import Optional
+
+from .navigator import NavInsID
+
+class NavigationScenarioData:
+    navigation: str
+    validation: str
+    pattern: str
+
+    def __init__(self, device: str, is_address_confirmation: bool, approve: bool):
```
xchapron-ledger, line 12: I'm not fan of `is_address_confirmation` an `enum` might be more future-proof? Or it could be introduced at a later time as this is not exposed to the apps.

src/ragger/navigator/navigation_scenario.py
```diff
@@ -0,0 +1,82 @@
...
+                self.pattern = "Confirm"
+            else:
+                self.navigation = NavInsID.USE_CASE_REVIEW_TAP
+                if approve:
+                    self.validation = [NavInsID.USE_CASE_REVIEW_CONFIRM]
+                else:
+                    self.validation = [NavInsID.USE_CASE_REVIEW_REJECT, NavInsID.USE_CASE_CHOICE_CONFIRM]
+                self.pattern = "Hold to sign"
+            self.validation += [NavInsID.USE_CASE_STATUS_DISMISS]
+
```
xchapron-ledger, line 34: Maybe add here an else raising an unimplemented error?

src/ragger/navigator/navigation_scenario.py
```diff
@@ -0,0 +1,82 @@
...
+class NavigationScenarioData:
+    navigation: str
+    validation: str
+    pattern: str
+
+    def __init__(self, device: str, is_address_confirmation: bool, approve: bool):
+        if device.startswith("nano"):
+            self.navigation = NavInsID.RIGHT_CLICK
+            self.validation = [NavInsID.BOTH_CLICK]
+            self.pattern = "Approve" if approve else "Reject"
```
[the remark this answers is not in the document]
fbeutin-ledger: I did not know we had regex, done

src/ragger/navigator/navigation_scenario.py
fbeutin-ledger: Done

src/ragger/navigator/navigation_scenario.py
fbeutin-ledger: Done

src/ragger/navigator/navigation_scenario.py
```diff
@@ -0,0 +1,115 @@
...
+                text=scenario.pattern,
+                path=path if path else self.screenshot_path,
+                test_case_name=test_name if test_name else self.test_name)
+        else:
+            self.navigator.navigate_until_text(
+                navigate_instruction=scenario.navigation,
+                validation_instructions=scenario.validation,
+                text=scenario.pattern,
+                path=path if path else self.screenshot_path,
+                test_case_name=test_name if test_name else self.test_name)
```
xchapron-ledger, line 83:
```suggestion
        if do_comparison:
            path = None
            test_case_name = None
        else:
            if not path:
                path = self.screenshot_path
            if not test_name:
                test_name = self.test_name
        self.navigator.navigate_until_text_and_compare(
            navigate_instruction=scenario.navigation,
            validation_instructions=scenario.validation,
            text=scenario.pattern,
            path=path,
            test_case_name=test_name)
```

fbeutin-ledger: Not a fan of the behavior of `navigate_until_text_and_compare` that does actually not compare

xchapron-ledger (commented):
LGTM

PR #181, opened 2024-04-05
Fix snapshot race in wait_for_text_on_screen for Speculos backend

fbeutin-ledger:


src/ragger/backend/speculos.py
```diff
@@ -241,6 +241,8 @@ def get_current_screen_content(self) -> dict:
         return self._retrieve_client_screen_content()
 
     def compare_screen_with_text(self, text: str) -> bool:
+        # Update self._last_screenshot to keep coherent internal state
+        self._last_screenshot = BytesIO(self._client.get_screenshot())
         for event in self._retrieve_client_screen_content()["events"]:
             if match(text, event.get("text", "")):
                 return True
```
xchapron-ledger, line 248: Shouldn't it be just before the return?

fbeutin-ledger: Not sure it changes anything, the `for` loop is only a few elements long
We want to be the closest to `_retrieve_client_screen_content`
I've considered putting it inside `_retrieve_client_screen_content` but I'm worried of side effects

xchapron-ledger: you could retrieved the events, then update the screenshot, and eventually loop on the event. WDYT?

src/ragger/backend/speculos.py
```diff
@@ -246,6 +247,29 @@ def compare_screen_with_text(self, text: str) -> bool:
...
+                                        text: str,
+                                        timeout: float = 10.0) -> None:
+        endtime = time() + timeout
+        # Only manual ticks sent by compare_screen_with_text in this function because
+        # we don't want a desync between screen and events
+        self.pause_ticker()
+        while True:
+            if self.compare_screen_with_text(text) == should_be_on_screen:
+                # TODO: investigate while this line is needed.. It should not be
+                self._last_screenshot = BytesIO(self._client.get_screenshot())
```
xchapron-ledger, line 261: Did you try putting it betwwen the `pause_ticker` and the `while True`?

fbeutin-ledger: Yep, does not work
also tried putting it between the `while True` and the `self.compare_screen_with_text`, does not work either

xchapron-ledger (approved):

PR #186, opened 2024-05-06
Adding the swipe gesture

lpascal-ledger:
- Adding proxy `finger_swipe` methods in the backends, only really implemented in Speculos for now (need >= v0.9.0)
   - GUI-specific implementations still need to be done (in another PR?)
- Added an `Center(Element)` (position-only class, not tied to a layout or a use case).
  The thing is, a "swipe" is a bit like a "touch", it is not attached to a specific layout or use case. Still, I feel like basic swipe should be usable from anywhere.
  So I added a basic swipe like the touch, but also 2 more specific instructions: `SWIPE_CENTER_TO_LEFT` and `SWIPE_CENTER_TO_RIGHT` which I hope will embed 90% (100%? :crossed_fingers: ) of current use cases.
- Changed the `UseCase.TX_REVIEW` scenario to use the swipe instead of the tap.
  - This may need some thinking, as this specific use case may be better having a "next" method rather than a "tap" one. And it would be our responsibility to implement either a swipe or a tap inside it, directly into the `firmare` module (it should be managed there afterall). But I don't had time to think of it more.

### Tested on 
- monero tests
 ```--- a/tests/monero_client/monero_cmd.py
+++ b/tests/monero_client/monero_cmd.py
@@ -229,7 +229,7 @@ class MoneroCmd(MoneroCryptoCmd):
             instructions = get_nano_review_instructions(1)
         else:
             instructions = [
-                NavIns(NavInsID.USE_CASE_REVIEW_TAP)
+                NavInsID.SWIPE_CENTER_TO_LEFT
             ]
```

- Ethereum tests - no changes as it is abstracted into the `UseCase.TX_REVIEW` scenario

### DIFF
--- a/src/ragger/navigator/navigation_scenario.py
+++ b/src/ragger/navigator/navigation_scenario.py
@@ -23,8 +23,12 @@ class NavigationScenarioData:
             self.pattern = "^Approve$" if approve else "^Reject$"
 
         elif device in [Firmware.STAX, Firmware.FLEX]:
-            if use_case == UseCase.ADDRESS_CONFIRMATION:
+            if device == Firmware.STAX:
                 self.navigation = NavInsID.USE_CASE_REVIEW_TAP
+            else:
+                self.navigation = NavInsID.SWIPE_CENTER_TO_LEFT
+
+            if use_case == UseCase.ADDRESS_CONFIRMATION:
                 if approve:
                     self.validation = [NavInsID.USE_CASE_ADDRESS_CONFIRMATION_CONFIRM]
                 else:
@@ -32,10 +36,6 @@ class NavigationScenarioData:
                 self.pattern = "^Confirm$"
 
             elif use_case == UseCase.TX_REVIEW:
-                if device == Firmware.STAX:
-                    self.navigation = NavInsID.USE_CASE_REVIEW_TAP
-                else:
-                    self.navigation = NavInsID.SWIPE_CENTER_TO_LEFT
                 if approve:
                     self.validation = [NavInsID.USE_CASE_REVIEW_CONFIRM]
                 else:
--- a/src/ragger/navigator/navigation_scenario.py
+++ b/src/ragger/navigator/navigation_scenario.py
@@ -31,7 +31,10 @@ class NavigationScenarioData:
                 self.pattern = "^Confirm$"
 
             elif use_case == UseCase.TX_REVIEW:
-                self.navigation = NavInsID.USE_CASE_REVIEW_TAP
+                if device.startswith("stax"):
+                    self.navigation = NavInsID.USE_CASE_REVIEW_TAP
+                else:
+                    self.navigation = NavInsID.SWIPE_CENTER_TO_LEFT
                 if approve:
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

**Review by cedelavergne-ledger — DISMISSED** [review]
Good for me, tested with app-ethereum

File: src/ragger/navigator/navigation_scenario.py [hunk 1]
```diff
@@ -31,7 +31,10 @@ def __init__(self, device: str, use_case: UseCase, approve: bool):
                 self.pattern = "^Confirm$"
 
             elif use_case == UseCase.TX_REVIEW:
-                self.navigation = NavInsID.USE_CASE_REVIEW_TAP
+                if device.startswith("stax"):
+                    self.navigation = NavInsID.USE_CASE_REVIEW_TAP
```
xchapron-ledger, line 35: No SWIPE for `ADDRESS_CONFIRMATION` above?

src/ragger/navigator/navigation_scenario.py
```diff
@@ -31,7 +31,10 @@ def __init__(self, device: str, use_case: UseCase, approve: bool):
                 self.pattern = "^Confirm$"
 
             elif use_case == UseCase.TX_REVIEW:
-                self.navigation = NavInsID.USE_CASE_REVIEW_TAP
+                if device.startswith("stax"):
```
xchapron-ledger, line 34:
```suggestion
                if device == "stax":
```
using `startswith` should probably only used on `nano`...

lpascal-ledger: Moreover I'm not sure in the first place why we are using strings when there is an enum.

src/ragger/navigator/navigation_scenario.py
lpascal-ledger: I'm not aware of all the cases where the SWIPE is mandatory; I get from your comment that it is for `ADDRESS_CONFIRMATION` so I'll add it.

xchapron-ledger: I'd say that swipe is the flex equivalent of tap on stax :)

lpascal-ledger: Okay, that will be simpler then!

xchapron-ledger (commented):
I'm wondering if we should create a `NavInsID.CONTINUE_NAVIGATION`
which would map on:
- `NavInsID.USE_CASE_REVIEW_TAP` for stax
- `NavInsID.SWIPE_CENTER_TO_RIGHT` for flex
- `NavInsID.RIGHT_CLICK` for nano

But maybe this will make things too opaque?

lpascal-ledger:
This PR highlighted a bug in Speculos which will be fixed by [this PR](https://github.com/LedgerHQ/speculos/pull/486).

lpascal-ledger:
@xchapron-ledger that may be interesting yes, but as for the implementation that would require adding / updates in the navigators, scenarios and firmware positions. This PR seems good enough as it is, I'd prefer having additional features in another one.

cedelavergne-ledger (approved):

PR #187, opened 2024-05-06
Adapting to new Speculos `finger_touch` API call.

lpascal-ledger:
Following [this PR](https://github.com/LedgerHQ/speculos/pull/481), the Speculos `finger_touch` function has two new arguments `x2` and `y2`.
We need to explicit that the third argument used in ragger is the `delay`, else it will implicitly understand that it is `x2`. This will break, as `x2` needs to be an integer, whereas the `delay` is often a float.


xchapron-ledger (dismissed):

lpascal-ledger:
Sorry @xchapron-ledger I needed to update one snapshot.

xchapron-ledger (approved):

PR #188, opened 2024-05-31
[add] Fixture to customize SpeculosClient arguments

lpascal-ledger:
This feature allows to inject custom parameters to the underlying `SpeculosClient` instance, allowing to further tweak its behavior (forcing a port for instance).

One would need to override the `additional_speculos_arguments` fixture in its `conftest.py` file:
```
import pytest
from typing import List
from ragger.conftest import configuration

pytest_plugins = ("ragger.conftest.base_conftest", )

@pytest.fixture(scope=configuration.OPTIONAL.BACKEND_SCOPE)
def additional_speculos_arguments() -> List[str]:
    return ["--api-port", "5000"]
```

This will work with a Speculos backend and only a Speculos backend.

If arguments are duplicated vs. the ones already computed and added by the Ragger fixtures (like the model, the display type, the seed, ...), the behavior is not specified. Be aware of it!

### DIFF
--- a/.github/workflows/fast-checks.yml
+++ b/.github/workflows/fast-checks.yml
@@ -63,3 +63,4 @@ jobs:
       with:
         builtin: clear,rare
         check_filenames: true
+        ignore_words_list: assertIn,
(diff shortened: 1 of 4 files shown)


bigspider (approved):
LGTM - tested with the test suite of the Bitcoin app.

PR #190, opened 2024-06-27
Navigation scenario depends on elf metadata

cedelavergne-ledger:
Update navigation_scenario manager to support conditional execution based on elf metadata:
- Swipe is introduced on Stax with API_LEVEL_21; need to support both `Tap to continue` and `Swipe to review`.
- NBGLv2 on Nano devices needs a new a validation step (Cf. https://github.com/LedgerHQ/app-boilerplate/pull/123)


cedelavergne-ledger:
Not needed

PR #192, opened 2024-07-08
Fix devices list from manifest

cedelavergne-ledger:
Read supported devices list in the manifest and only trigger the tests for the supported devices

Also fix a little typo when installing `click` package for the tests  (by the way, I don't know if this package is still needed :thinking: )


lpascal-ledger:
This `click` thingy is an old patch which I think aims at fixing dependency contradictions between `ledgerctl` (which needs `click`) and `blue` (or `speculos`?).
Indeed we may check if it is still needed.

PR #204, opened 2024-07-25
Fix when Rust app is a workspace

yogh333:


tdejoigny-ledger (commented):
could you please provide an example where we have the issue ?

src/ragger/utils/misc.py
```diff
@@ -72,7 +74,11 @@ def find_application(base_dir: Path, device: str, sdk: str) -> Path:
         if device == "nanos2":
             device = "nanosplus"
         app_name = toml.load(base_dir / "Cargo.toml")["package"]["name"]
-        app = app / "target" / device / "release" / app_name
+        cmd = ["cargo", "metadata", "--no-deps"]
+        output = subprocess.check_output(cmd)
+        metadata = json.loads(output)
+        target = Path(metadata["target_directory"])
+        app = target / device / "release" / app_name
```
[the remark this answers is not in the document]
yogh333: This fix enables to retrieve the absolute path to the binary (elf file) produced by `cargo build`. When the application repository is organised as a Rust workspace (several packages in the same crate), then the `target` folder is at the root of the crate and not in the application package folder. By using `cargo metadata`we are able to retrieve the proper path for all use cases (workspace or non workspace).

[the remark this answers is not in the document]
yogh333: As an example, you could check the Starknet app [repository](https://github.com/LedgerHQ/app-starknet/tree/y333/poseidon_support)

tdejoigny-ledger (dismissed):

tests/unit/utils/test_misc.py
```diff
@@ -40,13 +41,16 @@ def test_find_application_ok_c(self):
     def test_find_application_ok_rust(self):
         device, sdk, appname = "device", "rust", "rustapp"
         with temporary_directory() as dir_path:
-            tmp_dir = (dir_path / "target" / device / "release")
+            os.chdir(dir_path)
```
lpascal-ledger, line 44: Instead of doing `os.chdir(<path>); subprocess.check_output(cmd)`, you can just do `subprocess.check_output(cmd, cwd=<path>)`.

tests/unit/utils/test_misc.py
```diff
@@ -40,13 +41,16 @@ def test_find_application_ok_c(self):
...
+            cmd = ["cargo", "new", appname]
+            subprocess.check_output(cmd)
+            app_path = dir_path / appname
+            os.chdir(app_path)
```
lpascal-ledger, line 48: Is it necessary to `chdir` here? All path should be absolute.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.22.2] - 2024-07-25
+
+### Fixed
```
lpascal-ledger, line 10: This feels more like a change than a fix to me. It was working previously, it now needs to be adapted to new Rust build system behavior. No bug nor fix here.

lpascal-ledger (dismissed):

tests/unit/utils/test_misc.py
yogh333: Yes correct

lpascal-ledger (approved):

tests/unit/utils/test_misc.py
yogh333: Well actually it is mandatory for the test to pass as we must be in the Rust app root to use `cargo metadata` cmd.

lpascal-ledger: In that case it should rather be the Ragger code `output = subprocess.check_output(["cargo", "metadata", "--no-deps"])` which should explicit the `cwd` where it will be executed (`base_dir` I think?), rather than the whole system being in the correct directory when executing it; otherwise this piece of code would not be very robust.

lpascal-ledger (changes requested):

tests/unit/utils/test_misc.py
yogh333: Yes correct

lpascal-ledger:
@yogh333 I did not see the CI was failing due to `codespel`, if you want to silence it you can modify the `fast-checks.yml` last line to be like `ignore_words_list: assertIn,crate,`

PR #206, opened 2024-09-23
Correct position of choice 1 in stax choicelist

ajinkyaraj-23:
The position on stax for choicelist option 1 was incorrect. using choose(1) would not result in choice 1 but choice 2. Corrected it to right value.


lpascal-ledger:
Thanks @ajinkyaraj-23, indeed Ragger is not up to date on these positions.
We are no going to merge your PR because it only fixes the first choice, but we are going to update all the Stax/Flex choice list position very shortly.

lpascal-ledger:
Fixed in #208

PR #238, opened 2025-04-29
Use pytest builtin --log-level instead of custom --Verbose

fbeutin-ledger:


fbeutin-ledger:
I disagree with sonarcloud analysis

src/ragger/conftest/base_conftest.py
```diff
@@ -344,6 +339,15 @@ def pytest_collection_modifyitems(config, items):
                 pytest.mark.skip(reason=f"Test requires setup '{needed}' but setup is '{current}'"))
 
 
+# Fixture like function that will configure the ragger log level
+# Called by pytest_configure and not a fixture to ensure it always runs first
+def _setup_global_logging(config):
+    level_name = (config.getoption("--log-level") or "INFO").upper()
```
cedelavergne-ledger, line 345: Please use `DEBUG` level by default, which avoid to always run with additional option. Same below...

src/ragger/backend/ledgerwallet.py
```diff
@@ -91,7 +91,7 @@ def receive(self) -> RAPDU:
 
     @raise_policy_enforcer
     def exchange_raw(self, data: bytes = b"", tick_timeout: int = 0) -> RAPDU:
-        self.apdu_logger.debug("=> %s", data.hex())
+        self.apdu_logger.info("=> %s", data.hex())
```
cedelavergne-ledger, line 94: Not convinced at all by all these new logs that will polluate in most cases...

fbeutin-ledger: This is the APDUs exchanged.
I can't see a scenario where it's useful to look at the logs while not being interested in the apdus exchanged

src/ragger/backend/speculos.py
```diff
@@ -265,8 +265,12 @@ def get_current_screen_content(self) -> dict:
         return self._retrieve_client_screen_content()
 
     def compare_screen_with_text(self, text: str) -> bool:
+        logger = get_default_logger()
+        logger.debug(f"Looking for pattern '{text}' in current screen")
         for event in self._retrieve_client_screen_content()["events"]:
-            if match(text, event.get("text", "")):
+            event_text = event.get("text", "")
+            logger.debug(f"Candidate event text '{event_text}'")
```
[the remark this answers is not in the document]
fbeutin-ledger: It's indeed only for debugging ragger and not at all intended for ragger users
That's why it's behind the `debug` level

src/ragger/conftest/base_conftest.py
fbeutin-ledger: `DEBUG` contains logs for debugging ragger / navigation / pattern matching and so on
Users should never need debug logs

cedelavergne-ledger (approved):

cedelavergne-ledger (approved):

PR #245, opened 2025-06-10
minor bug fixes on physical backend

mruiz-ledger:


iartemov-ledger (commented):
Looks nice and not intrusive! 
I'd like at first to test it locally with  `app-bitcoin-new`. 
There are few apps that use `exchange_async()` https://github.com/search?q=org%3ALedgerHQ+exchange_async+language%3APython&type=code&l=Python, so interesting to see the impact.

iartemov-ledger (approved):

PR #249, opened 2025-06-27
hotfix: yield true on async to notify the action has been completed

mruiz-ledger:
In reality, the apdu exchange might not be completed, but the physical device is going to block either way till the flow has been done manually

### DIFF
--- a/src/ragger/backend/ledgercomm.py
+++ b/src/ragger/backend/ledgercomm.py
@@ -106,7 +106,7 @@ class LedgerCommBackend(PhysicalBackend):
         return result
 
     @contextmanager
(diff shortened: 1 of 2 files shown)


iartemov-ledger (approved):
Tested with different combinations (flex/nanosp, ledgercomm/ledgerwallet, --display or not) of directly impacted app (`app-bitcoin-new` and `app-bitcoin`) and with the ones using `exchange_async_raw()` (`app-ethereum` and `app-monero`).  Everything is OK. 
No UI guidance with `app-bitcoin-new` and `app-bitcoin` as expected, and there is with `app-ethereum` and `app-monero` when `--display` option is passed.

PR #254, opened 2025-07-17
fix: correct typos in README and interface.py documentation

arcaus:
Fixed some minor typos in the codebase as well as the README


src/ragger/backend/interface.py
```diff
@@ -205,7 +205,7 @@ def exchange(self,
         :type ins: int
         :param p1: First instruction parameter, defaults to 0
         :type p1: int
-        :param p1: Second instruction parameter, defaults to 0
+        :param p2: Second instruction parameter, defaults to 0
```
cedelavergne-ledger, line 208: Agree, but please update also the followng line :wink: 
Same for other comments....

cedelavergne-ledger (approved):

PR #273, opened 2026-04-27
Force display to be set when no-nav is on the commandline

cedelavergne-ledger:
Bonus:

Add 2 new markers to ease test skip:
- `skip_on_backend(backend)` -> This is the reverse of the existing `use_only_on_backend(backend)`
- `skip_nano`

### DIFF
--- a/src/ragger/conftest/base_conftest.py
+++ b/src/ragger/conftest/base_conftest.py
@@ -81,7 +81,7 @@ def backend_name(pytestconfig):
 
 @pytest.fixture(scope="session")
 def display(pytestconfig):
-    return pytestconfig.getoption("display")
+    return pytestconfig.getoption("display") or pytestconfig.getoption("no_nav")
 
 
 @pytest.fixture(scope="session")
--- a/src/ragger/conftest/base_conftest.py
+++ b/src/ragger/conftest/base_conftest.py
@@ -502,6 +502,20 @@ def auto_skip_on_backend(request, backend_name):
             pytest.skip(f'⚠️ Not supported on {excluded_backend} backend')
 
 
+@pytest.fixture(autouse=True)
+def auto_skip_nano(request):
+    """Auto-skip tests marked with skip_nano when running on Nano devices."""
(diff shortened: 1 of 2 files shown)


src/ragger/conftest/base_conftest.py
```diff
@@ -492,6 +492,28 @@ def use_only_on_backend(request, backend_name):
             pytest.skip(f'skipped on this backend: "{current_backend}"')
 
 
+@pytest.fixture(autouse=True)
+def auto_skip_on_backend(request, backend_name):
```
fbeutin-ledger, line 496: We already have `use_only_on_backend` fixture
I think we can either merge or remove old one or adapt new one

src/ragger/conftest/base_conftest.py
```diff
@@ -492,6 +492,28 @@ def use_only_on_backend(request, backend_name):
...
+
+@pytest.fixture(autouse=True)
+def auto_skip_nano(request):
+    """Auto-skip tests marked with skip_nano when running on Nano devices."""
+    if request.node.get_closest_marker('skip_nano'):
+        try:
+            device = request.getfixturevalue('device')
+            if device.is_nano:
+                pytest.skip("⚠️ Not yet supported on Nano devices")
+        except pytest.FixtureLookupError:
```
fbeutin-ledger, line 513: Why except here? Needs a comment

cedelavergne-ledger: Just in case of the `device` is not in the commandline...

cedelavergne-ledger: Fixed, adding a comment
This is needed, because ragger tests doesn't use the `device` parameter

src/ragger/conftest/base_conftest.py
cedelavergne-ledger: I prefer to keep separate fixtures, it is easier to read and maintain

fbeutin-ledger (approved):

PR #274, opened 2026-04-29
Fix stack consumption fixture when test does not interact with device

mbrousset-ledger:


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.45.1] - 2026-04-27
```
bigspider:
```suggestion
## [1.45.2] - 2026-04-29
```

src/ragger/conftest/base_conftest.py
```diff
@@ -130,8 +130,12 @@ def get_stack_consumption(pytestconfig):
 @pytest.fixture(autouse=True)
 def stack_consumption_hooks(request, get_stack_consumption: bool):
     if get_stack_consumption:
-        backend = request.getfixturevalue("backend")
-        _backend_name = request.getfixturevalue("backend_name")
+        try:
+            backend = request.getfixturevalue("backend")
+            _backend_name = request.getfixturevalue("backend_name")
+        except pytest.FixtureLookupError:
+            yield
```
bigspider, line 149: Why does it need to `yield` here?

src/ragger/conftest/base_conftest.py
```diff
@@ -130,8 +130,12 @@ def get_stack_consumption(pytestconfig):
 @pytest.fixture(autouse=True)
 def stack_consumption_hooks(request, get_stack_consumption: bool):
     if get_stack_consumption:
```
bigspider: This might help future reviewers

```suggestion
    # This fixture is marked `autouse=True` so that it happens automatically
    # on tests that interact with the device. However, it should do nothing if
    # running tests that _do not_ interact with the device.
    # We detect that by verifying if the `backend` fixture is present.

    if get_stack_consumption:
```

bigspider (changes requested):
LGTM, with some nits

src/ragger/conftest/base_conftest.py
fbeutin-ledger: Duplicate try/except blocks:
The [request.getfixturevalue()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) calls are wrapped in identical try/except blocks both before and after yield.
This is unnecessary — generator locals persist across yield, so [backend](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [_backend_name](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) from before the yield can be reused after it.

mbrousset-ledger: Removed duplicate try/except blocks

bigspider (approved):

PR #279, opened 2026-06-05
Add --log_apdus option

fbeutin-ledger:


doc/tutorial_conftest.rst
```diff
@@ -480,12 +480,24 @@ will store the captured snapshots, with the ``--golden_run`` CLI argument.
...
 
-Recording the APDUs (``--log_apdu_file``)
-'''''''''''''''''''''''''''''''''''''''''
+Recording the APDUs (``--log_apdu_file`` / ``--log_apdus``)
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
 
 It can be useful to record all the APDU transmitted between the client and the
-application during a test. the ``--log_apdu_file`` allows to specify a file
-path in which every :term:`APDU` and :term:`RAPDU` will be recorded.
+application during a test.
```
cedelavergne-ledger, line 487: I think the default `--log_apdu_file` alone should be sufficient. Please double check with the option itself, without giving a file name...

